Issue #46, opened 2014-11-26
if bug


Farcek:
always else tag runing
/*****************************/

{% if(v == 1)%}
    ...
{% elseif(v == 2)%}
   ...
{% elseif(v==3)%}
  ...
{% else %}
... // the always running
{% endif%}

nhylated:
Hi,

This issue is reproducible with float and doubles and sometimes with integers as well. The issue was due to == being used for comparing two Number objects. The underlying issue is the one discussed [here](http://stackoverflow.com/questions/12226757/).

I've created pull request #53 for this along with a regression test. Let me know if you need anything else in the pull request.

Thanks.

mbosecke:
Resolved by pull request #53

PR #53, opened 2015-01-11
issue-46: Fix equals comparision of Number objects to use equals().

nhylated:
Issue #46 was occurring since equals comparision of Number objects was done using == which doesn't work in all cases, especially with Float and Double. See http://stackoverflow.com/questions/12226757/

Fixed this by using equals() for eqaulity comaprision of Number objects. Also added a regression test in CoreOperatorsTest.java.

See: https://github.com/mbosecke/pebble/issues/46


mbosecke:
I'm amazed this went unnoticed for so long. Thanks very much @nhylated.

nhylated:
You're welcome! :+1:

PR #56, opened 2015-03-09
NPE in GetAttributeExpression (caused by member cache)

UshakovVasilii:


UshakovVasilii:
Problem is reproduced after https://github.com/mbosecke/pebble/commit/76dae61254ec0b4a8a9f9963be47602581b630c6

NPE is occurred in `invokeMember` method on https://github.com/mbosecke/pebble/blob/v1.4.3/src/main/java/com/mitchellbosecke/pebble/node/expression/GetAttributeExpression.java#L135
Because `member` is not null (cached) and `object` is null

I cannot reproduce problem without `macro`

mbosecke:
Thanks! And sorry for taking so long to merge this one.

PR #60, opened 2015-04-04
fixed platfrom dependent issues

rseiler:
some tests were broken on my machine, because of other default locale and number formatting settings. i fixed the issues with this pull requests - all tests are now green on my machine.


mbosecke:
Thanks! This was a great contribution; something I never would've realized but probably affected others from contributing.

PR #62, opened 2015-04-05
Fix member cache for different types

UshakovVasilii:
We can pass different classes in same template. In this case [invokeMember](https://github.com/mbosecke/pebble/blob/v1.4.4/src/main/java/com/mitchellbosecke/pebble/node/expression/GetAttributeExpression.java#L155) will not work


mbosecke:
I knew this was a problem from the very first moment I wrote that bit of code but I could not fathom a scenario in which someone would pass different types to the same template so I ignored it until now. Because you noticed the issue, I was obviously wrong.

Thanks for the contribution.

PR #74, opened 2015-05-29
Add missing break statement in EscapeFilter

samlanning:
This missing break statement means it is impossible to use the escape strategy `url_param`, as it is treated as urecognized. This PR fixes this.


mbosecke:
Thanks! It was a silly mistake.

Issue #76, opened 2015-06-19
Refering to non-existing variable gives NPE


ethlo:
Version 1.5.0

When referring to a non-existent property in strict mode, object is null and throws NPE:

StackTrace:
```
java.lang.NullPointerException
	at com.mitchellbosecke.pebble.node.expression.GetAttributeExpression.evaluate(GetAttributeExpression.java:138)
	at com.mitchellbosecke.pebble.node.expression.FilterExpression.evaluate(FilterExpression.java:54)
	at com.mitchellbosecke.pebble.node.PrintNode.render(PrintNode.java:32)
	at com.mitchellbosecke.pebble.node.BodyNode.render(BodyNode.java:46)
	at com.mitchellbosecke.pebble.node.RootNode.render(RootNode.java:32)
	at com.mitchellbosecke.pebble.template.PebbleTemplateImpl.buildContent(PebbleTemplateImpl.java:61)
	at com.mitchellbosecke.pebble.template.PebbleTemplateImpl.evaluate(PebbleTemplateImpl.java:107)
	at com.mitchellbosecke.pebble.template.PebbleTemplateImpl.evaluate(PebbleTemplateImpl.java:88)
```

Problematic code in GetAttributeExpression.java:
```java
throw new AttributeNotFoundException(
                    null,
                    String.format(
                            "Attribute [%s] of [%s] does not exist or can not be accessed and strict variables is set to true.",
                            attributeName, object.getClass().getName()));
```

mbosecke:
Fixed by 1523d17. Thanks again.

PR #79, opened 2015-06-25
Fix for Pebble issue #76

ethlo:
I created RootAttributeNotFoundException as a subclasss of AttributeNotFoundException so that it is possible to distinguish the difference if need be. I also changed an existing test that expected NPE, as I feel this is a very generic exception that can hide other issues.


mbosecke:
I love it! Thanks a lot.

ethlo:
My pleasure. Will you cut a 1.5.1 release anytime soon? :)

mbosecke:
Absolutely, either tonight or tomorrow night for sure.

PR #83, opened 2015-07-30
Add "rsort" filter and fix primitive array subscripting.

Barakat:
Add "rsort" filter to sort list items in the reverse order.

Update: I've noted the comment in the on the ignored test `CoreFiltersTest#testSliceWithPrimitivesArray`. The issue is located in `GetAttributeExpression#evaluate`. The initial implementation handles only array of references, in the second commit I modified it to handle arrays of any type and removed `@Ignore` so now all tests pass.


mbosecke:
Thanks!

PR #112, opened 2015-11-10
Add a size function to return the size of a collection or map

ebussieres:
I just added a function to return the size of a collection or map


ebussieres:
I just saw that there's another pull request to add a length filter which does basically the same thing. Feel free to reject it if you wish :)

thomashunziker:
I was thinking also to call it size. However the original twig implementation uses the word 'length' hence I used the 'length' key word.

PR #116, opened 2015-11-11
Add support for parsing template in parallel.

thomashunziker:
In this pull request I have add support for parsing templates in parallel. The semaphore is not required anymore.

To test my changes I have add a unit test which I run before I change anything and with removed semaphore. The test fails as expected. The test does now not fail anymore after my changes.

I had to add some breaking changes:
* The lexer is not configurable anymore. The configuration is now in the separate class 'Syntax'. The syntax is immutable to make sure the syntax is not changed in the middle of the parsing.
* The NodeVisitors are not thread safe. Hence I had to add the factory pattern. So the visitors are created for each template parsing. This makes the whole thing thread-safe. The real change is that the extensions now needs to provide NodeVisitorFactory objects instead of NodeVisitor objects.
* I create for each template parsing a new Lexer and Parser object. This should not have any performance impact since I refactor the handling of the Syntax, which compiles the regex.
* The TokenParser needs to pass the parser as argument.

I reviewed the rest of the code. It seems as these are the only changes which are required to make the whole thing thread-safe. I found some spots which eventually can cause problems, however they exists before and have no effect on the parsing. They have only an effect on the rendering (see the new TODOs). 

If you see any problem please advise me!

### DIFF
--- a/src/main/java/com/mitchellbosecke/pebble/PebbleEngine.java
+++ b/src/main/java/com/mitchellbosecke/pebble/PebbleEngine.java
@@ -19,7 +19,6 @@ import java.util.Map;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ExecutorService;
-import java.util.concurrent.Semaphore;
 
 import com.google.common.cache.Cache;
 import com.google.common.cache.CacheBuilder;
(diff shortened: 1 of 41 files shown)


mbosecke:
Looks great! It does have an affect on backwards-compatibility but that's okay because it's about time I release a version 2.0. This does, however, mean I want to add a few more changes before I release the next version.

thomashunziker:
Its definitely a good idea to put this change into a version 2.x. We currently rolling out Pebble in our application. Eventually we come across other issues. So I will potentially create more pull requests.

PR #117, opened 2015-11-11
Added a cache tag.

ebussieres:
You can cache the rendering of a portion of a page by using it. Very
useful for static part of a website (a menu/footer for example). It uses
the name and the locale as a key in the cache.

I got a lot of performance improvement by doing so. Our menu/footer uses a lot of function to be constructed and comes from the database, but once constructed, it basically stays the same.

Usage example :
```javascript
{% cache menu %}
    {% for item in items %}
        {{ item.text }}
        ....
    {% endfor %}
{% endcache %}
```

### DIFF
--- a/src/main/java/com/mitchellbosecke/pebble/node/CacheNode.java
+++ b/src/main/java/com/mitchellbosecke/pebble/node/CacheNode.java
@@ -11,22 +11,103 @@ package com.mitchellbosecke.pebble.node;
 import java.io.IOException;
 import java.io.StringWriter;
 import java.io.Writer;
+import java.util.Locale;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutionException;
 
 import com.google.common.cache.Cache;
+import com.mitchellbosecke.pebble.cache.BaseTagCacheKey;
 import com.mitchellbosecke.pebble.error.PebbleException;
 import com.mitchellbosecke.pebble.extension.NodeVisitor;
 import com.mitchellbosecke.pebble.template.EvaluationContext;
 import com.mitchellbosecke.pebble.template.PebbleTemplateImpl;
+import com.mitchellbosecke.pebble.tokenParser.CacheTokenParser;
 
+/**
+ * Node for the cache tag
+ *
+ * @author Eric Bussieres
+ */
 public class CacheNode extends AbstractRenderableNode {
+    /**
+     * Key to be used in the cache
+     *
+     * @author Eric Bussieres
+     */
+    private class CacheKey extends BaseTagCacheKey {
+        private final Locale locale;
+        private final String name;
+
+        public CacheKey(String name, Locale locale) {
+            super(CacheTokenParser.TAG_NAME);
+            this.name = name;
+            this.locale = locale;
+        }
+
+        /**
+         * {@inheritDoc}
+         *
+         * @see java.lang.Object#equals(java.lang.Object)
+         */
+        @Override
+        public boolean equals(Object obj) {
+            if (this == obj) {
+                return true;
+            }
+            if (!super.equals(obj)) {
+                return false;
+            }
+            if (this.getClass() != obj.getClass()) {
+                return false;
+            }
+            CacheKey other = (CacheKey) obj;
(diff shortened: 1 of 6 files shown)


src/main/java/com/mitchellbosecke/pebble/node/CacheNode.java
```diff
@@ -0,0 +1,73 @@
...
+import com.google.common.cache.CacheBuilder;
+import com.mitchellbosecke.pebble.error.PebbleException;
+import com.mitchellbosecke.pebble.extension.NodeVisitor;
+import com.mitchellbosecke.pebble.template.EvaluationContext;
+import com.mitchellbosecke.pebble.template.PebbleTemplateImpl;
+
+public class CacheNode extends AbstractRenderableNode {
+
+    private final BodyNode body;
+    private final Cache<String, String> cache;
```
jcarvalho: Isn't this `Cache` instance unique to each `cache` node? A different instance of `CacheNode` is created for different usages, meaning that this cache in practice is only keeping the multiple locales in which the template is rendered.

jcarvalho:
Shouldn't a mechanism to invalidate the caches be provided?

Say that you are caching a page's footer with semi-static information. It should be possible for my code to either explicitly say 'Hey, this information has been changed, please purge your cache', or have the cache expire after a little while, so that the new information may be rendered.

src/main/java/com/mitchellbosecke/pebble/node/CacheNode.java
ebussieres: No, when the pebble cache is active, only one instance of the cache node exists

jcarvalho: That only applies to the process of compiling the text contents into the Node Tree.

For every instance of `{% cache foo %}` you have in your template, you will have an instance of `CacheNode`, otherwise the `body` and `name` variables would be shared among all cache blocks :). I believe @mbosecke can back me up on this one.

ebussieres: Why would you want to have multiple {% cache foo %} instance ? It doesn't make sense. Each cache has its name

ebussieres:
I'll make some change to configure the cache in PebbleEngine and pass the reference to the cacheNode. So you'll be able to initialize it with a TTL if you wish or reset the cache.

src/main/java/com/mitchellbosecke/pebble/node/CacheNode.java
ebussieres: Anyway i've just commited to configure the cache in PebbleEngine and pass the reference to the cacheNode, so there will be only one instance of the cache.

jcarvalho: Say I want to cache some part of the header of my page, and some part of the footer. These will be two distinct `cache` tags. This will generate two different `CacheNode` instances in the tree.

ebussieres: You're right I understand what you're saying. Just check my last commit I think it would be better this way ;)

thomashunziker:
I'm not so sure if it is a good idea to make a cache per tag. I think a better way would be to use one cache for all tags. 

Additionally you build the cache key by combining two strings
```
String key = this.name + context.getLocale();
```
You should better use a dedicated object which correctly implements the equals and hashCode methods.

ebussieres:
Nice ideas, i'll make changes this week

ebussieres:
I've just made changes as @thomashunziker suggested.

pacey:
I just wanted to say this is great feature! Really looking forward to using this. Good work guys :+1:

mbosecke:
Very cool feature :+1:

donnior:
How about changing cache tag's name from static string to expression, we can get it more powerful. 

Like :
```
{% cache 'item' + item.id %}

... item detail html content

{% endcache %}
```

I've changed some code and tested it on my machine without expire strategy, seems everything is ok.  
Could you please think about this feature?

@ebussieres @mbosecke

ebussieres:
@donnior  I've just made a pull request #141 for this feature

PR #120, opened 2015-11-13
Enum comparision

thomashunziker:
Within the equals expression I add support to compare enums with strings. E.g. the following works now:

```
{% if my.type == 'MY_ENUM_NAME' %}
```


ebussieres:
This expression doesn't work when accessing a map with an enum as a key. Should be in GetAttributeExpression I think

```javascript
{{ mymap.get('ENUM_ENTRY') }}
```

PR #121, opened 2015-11-13
Add isolate key word to 'include'

thomashunziker:
In certain situation it is required that the include is executed in isolation. Meaning the included code should be evaluated in a dedicated scope.

### DIFF
--- a/src/main/java/com/mitchellbosecke/pebble/PebbleEngine.java
+++ b/src/main/java/com/mitchellbosecke/pebble/PebbleEngine.java
@@ -62,7 +62,7 @@ public class PebbleEngine {
     /*
      * Major components
      */
-    private Loader loader;
+    private Loader<?> loader;
 
     private final Parser parser;
 
@@ -80,7 +80,7 @@ public class PebbleEngine {
     /**
      * Template cache
      */
-    private Cache<String, PebbleTemplate> templateCache;
+    private Cache<Object, PebbleTemplate> templateCache;
 
     /*
      * Extensions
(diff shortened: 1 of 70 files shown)


mbosecke:
The more I look at this the more I think that it should be "isolated" by default and a user would have to opt-out of the isolation if they want to.

One reason I propose to have it isolated by default is because it would feel like a bug if you accidentally overwrite variables that exist later in the current template. For example:

    {% set 'foo' = 'bar' %}
    {% include 'header' with { 'foo' : 'baz' } %}
    {{ foo }} 

Intuitively I think the last line of this template should print "bar" instead of "baz". Ultimately, if it isn't isolated by default then the "with" statement is really syntactical sugar for the "set" statement. 

In fact, I don't even know why any one would opt-out of the isolation. I recommend we just always have it isolated; what do you think?

mbosecke:
If you're on board, I can just merge your first commit, 01ca4d2, which I think will do the trick.

thomashunziker:
Personally I totally agree with you. However when I interpret the code correctly the isolation was already not given before the change with the key word 'with'. The included template was always capable of changing the outer variables. The question is if we need to be backward compatible? Eventually we can add a key word which allows to by pass the isolation.

mbosecke:
I have committed 01ca4d2. The next release is version 2.0 so I'm okay with breaking some backwards compatibility at this point and I can't yet think of a use case where isolation is bad. If I do, I'll revisit this task and probably provide a way for people to opt-out of the isolation.

PR #122, opened 2015-11-14
Iterate over a sequence of numbers

ebussieres:
I have implemented the range function to iterate over a sequence of number

```javascript
{% for i in range(0, 3) %}
    {{ i }},
{% endfor %}

{# outputs 0, 1, 2, 3, #}
```

When increment is given (as the third parameter), it specifies the increment (or decrement):

```javascript
{% for i in range(0, 6, 2) %}
    {{ i }},
{% endfor %}

{# outputs 0, 2, 4, 6, #}
```

.. operator is just a shortcut for the range function with an increment of 1

```javascript
{% for i in 0..5 %}
    {{i}}
{% endfor %}

{# outputs 012345 #}
```

You can achieve exactly the same thing by iterating over character as well :
```javascript
{% for i in 'a'..'e' %}
    {{i}}
{% endfor %}

{# outputs abcde #}
```

### DIFF
--- a/src/main/java/com/mitchellbosecke/pebble/extension/core/CoreExtension.java
+++ b/src/main/java/com/mitchellbosecke/pebble/extension/core/CoreExtension.java
@@ -163,6 +163,7 @@ public class CoreExtension extends AbstractExtension {
 
         functions.put("max", new MaxFunction());
         functions.put("min", new MinFunction());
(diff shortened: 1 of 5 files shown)


mbosecke:
Really great stuff, thanks a lot!

PR #126, opened 2015-11-16
Add supports for BigDecimal when comparing Number

ebussieres:


r0b0tbuilder:
:octocat: Just another idea here, but what about comparing anything that implements `java.lang.Comparable`?  This implementation would support comparing of many data types, including `BigDecimal`.

ebussieres:
I don't think java.lang.Number implements Comparable, so we have to cast it.

r0b0tbuilder:
Although this could easily be a separate issue, it does affect your code dealing with the comparison if `BigInteger`s.  My suggestion is that the comparison methods in `OperatorUtils` be re-implemented so that they that use `java.lang.Comparable` to ensure that all number data types (`int`s, `float`s, etc...) can be easily compared.

```java
    public static boolean gt(Comparable<?> op1, Object op2) {
        return op1.compareTo(op2) > 0;
    }

    public static boolean gte(Comparable<?> op1, Object op2) {
        return op1.compareTo(op2) >= 0;
    }

    public static boolean lt(Comparable<?> op1, Object op2) {
        return op1.compareTo(op2) < 0;
    }

    public static boolean lte(Comparable<?> op1, Object op2) {
        return op1.compareTo(op2) <= 0;
    }

```

In the above example there is no need for method implementations to compare objects (like `wideningConversionBinaryOperation`), when `Comparable<T>` can be used instead.

Does that make sense?

ebussieres:
I've just refactored and simplified all comparisons and operations to support all type (BigDecimal, Double, Float, BigInteger, Long, Integer, Short, Byte).

mbosecke:
This is great; I appreciate the work you put into this. I merged it in but not without a couple of modifications/notes:

- I'm hesitant about the code that parses numbers from strings to perform binary comparisons, ex. "2"<"4". I can't think of a very good use case and it's arguable that string comparisons should actually perform a letter-by-letter alpha comparison, i.e. "abc" < "abd", which can be helpful for sorting. Regardless, I kept that code in for now but I'll probably leave it undocumented until someone brings up a good use case.
- I removed the code that attempts to perform binary operations on strings, ex. "2" + "3". I did so because a "+" operator is reserved for string concatenation. So "2"+"2" should equal "22", not "4". Even if we kept the "+" operator as a concatenation operator (which you did) I still wouldn't want the other operators implemented for strings because then there would be an expectation that addition would be available too.
- I added back the "integerOperation" method which was removed from the OperatorUtils and replaced with "longOperation" even if both operands are integers. All of our tests pass without it but I still want to strictly follow the Java specification because someone out there may be relying on the fact that two integers added together will return an integer. Perhaps someone is even relying on integer overflow/underflow.

ebussieres:
I agree with all the changes you made. We can also remove the string comparison, I have no problem with it. And I would replace it with a filter to parse a string as a Number. What do you think about it ?

ebussieres:
@mbosecke As for the third point, if somebody rely on overflow/underflow, we should also add Short and Byte operation ?

mbosecke:
I had to double check this, and I hope I'm reading it correctly, but according to the java spec it looks like chars, shorts, and bytes are converted to integers during a binary operation:

https://docs.oracle.com/javase/specs/jls/se8/html/jls-5.html#jls-5.6.2

PR #132, opened 2015-11-22
Added index0 for Twig compatiliblity 

babs:
Just to add some compatibility between pebble and twig:

Twig index is 1 based :/ The 0 based index is called index0
(historical issue)

### DIFF
--- a/src/main/java/com/mitchellbosecke/pebble/node/ForNode.java
+++ b/src/main/java/com/mitchellbosecke/pebble/node/ForNode.java
@@ -92,8 +92,9 @@ public class ForNode extends AbstractRenderableNode {
                 loop.put("last", index == length - 1);
                 loop.put("first", index == 0);
                 loop.put("revindex", length - index - 1);
+                loop.put("revindex0", length - index - 1);
+                loop.put("index0", index);
                 loop.put("index", index++);
-                loop.put("length", length);
                 loop.put("length", length);
 
                 context.put("loop", loop);


ebussieres:
Actually your code won't work as index0 and index would be the same.

You should do this instead. Am I right ?

```java
loop.put("index0", index);
loop.put("index", ++index);
```

babs:
No, this is intended but I didn't made myself clear :)
pebble's loop.index is zero base but twig is 1 based.
To circumvent the issue, twig added a index0.
My suggestion is to add a compatible statement explicitly named index0.
In the other side, I made a pull request on Twig side to also add a index1 and revindex1 to have both explicit (0 and 1 indices) and the implicit one more like implementation dependent :)

ebussieres:
So index0 and index are the same ? The purpose is just to be like twig ?

babs:
sorry I edited before I saw your second reply.
My intention was to make it compatible without breaking current uses of loop.index :)

derekvandaal:
Hi guys

If I may chime in here. In terms of pebble core. Shouldn't the design
philosophy be as pebble is pebble and it isn't necessarily directly
compatible with any other template language?

I for example started using pebble because I worked with python's Django
before and fell in love with its templating language.

Pebble is already so extensible and unless a change adds real value, I
would vote to keep the core concise and simple? :)

ebussieres:
I agree it isn't a copy of twig but inspired. It doesn't add anything new.

babs:
I totally get it.
I just thought that making it compatible with twig, as it doesn't require a huge change nor mess up that much with the code, would be a really good point that worthed a try.

ebussieres:
Just let @mbosecke decide :)

babs:
@ebussieres Your patch is perfectly fine and ensure a complete compatibility with Twig and Jinja if you're not afraid of altering the index behavior.

mbosecke:
I agree with the sentiment that the overall design philosophy of Pebble is that it is inspired by Twig and Django but will deviate in an attempt to be better. 

Saying that, I see where you are coming from, @babs, and I appreciate the pull request. My biggest concern with this change is the sheer amount of conversations I will have from this point on explaining to people why there are two "index" variables that have the same value. I don't think the compatibility factor alone adds enough value to outweigh the potential confusion.

babs:
Ok thanks.
I think I'll go for a filter instead like norm_index where its behavior will change based on the implementation then.

PR #133, opened 2015-11-22
add 'divisible by' test support

donnior:
`divisible by` checks if a variable is divisible by a number; it also
has alias `divisibleby`

```
{% if loop.index is divisible by(3) %}
    ...
{% endif %}
```


ebussieres:
Why don't you use the modulo comparison ?

```javascript
{% if loop.index % 3 == 0 %}
    ...
{% endif %}
```

donnior:
Just found Twig has a built-in `divisible by` support, so I committed this PR. Now I will close it.

PR #136, opened 2015-11-23
Added 'array' type support for `last` and `first` filter

donnior:
Currently the `first` and `last` filter only support `String` and
`Collection` input, it should support `Object[]`


mbosecke:
Looks great, thanks

PR #139, opened 2015-11-25
Add better error handling within expression

thomashunziker:
I have add a method for getting the line number on the expression interface.

This helps to pass the line number along the exceptions. One thing which I do not like is the way I had to handle BinaryExpression and UnaryExpression. Since we are using reflections to create new instances we can not pass along the line number in the constructor. Eventually we should apply here the factory pattern. Instead we pass the class name in BinaryOperatorImpl resp. UnaryOperatorImpl we can pass a factory object. This factory object has a method to create new instances of the expressions. We could also pass easily the line number along the factory.

@mbosecke if you like to go the way with the factory I can change this. However the change with the factory would be an even bigger change. I would need to change essentially all expressions to also provide a factory class. In theory the implementation would be better since the line number can be final and we do not need reflections which is typically harder to understand for new developers.


mbosecke:
This is a really good change; I really like the idea of the line number on the expression interface. Thanks!

And regarding the usage of a factory pattern to construct expressions, I'm definitely interested in this idea. It would be quite a big change but I think it would be for the better and I would accept a pull request for this.

Issue #127, opened 2015-11-17
improvement: retrieve current locale used


jmsanzg:
Sometimes the translations are not in a resource bundle but in the beans themselves. For instance, imagine an Article which may have a name translated to multiple languages. The best way to handle this would be to have the locale parameter available directly on the template.

ebussieres:
If you are using pebble-spring3, I have just maded a pull request (https://github.com/mbosecke/pebble-spring3/pull/5) to access the current locale.

jmsanzg:
@ebussieres that looks amazing!. Sadly I'm using Pebble calling it directly to generate an email from a template. I'll check your commit and I can try to pull the same functionality for no-spring use.

ebussieres:
You can just create a simple function like this

```java
public class LocaleFunction implements Function {
    @Override
    public Object execute(Map<String, Object> args) {
        EvaluationContext context = (EvaluationContext) args.get("_context");
        return context.getLocale();
    }

    @Override
    public List<String> getArgumentNames() {
        return null;
    }
}
```

jmsanzg:
I'll use that function and, if I can vote, I'd put it on the core list of available functions.

mbosecke:
Added a global variable called "locale" in commit 4862f9d. It can be accessed just like any other user-provided variable:

```
<html>
    <body>
        <p>The current locale is: {{ locale }}</p>
    </body>
</html>
```

PR #142, opened 2015-11-29
PR for issue #127: Make current locale accessible

ebussieres:
Pull request for #127.

If it's accepted, i'll remove the locale variable in my pebble-spring3 pull request (https://github.com/mbosecke/pebble-spring3/pull/5)


mbosecke:
Thanks for the pull request. I didn't merge this directly but I just pushed commit #4862f9d which is pretty much exactly the same with two small differences:
- "locale" variable is inserted into context before the global variables from the user-provided extensions. This allows users to override it for whatever reason.
- Also added a "template" global variable which references the template itself. Users can use this to get the template name via "template.name" which has been requested in the past.

PR #144, opened 2015-12-01
Fix for bean method with Null Arguments 

ebussieres:
When passing null as an argument, a NullPointerException is thrown. This
PR fixes it


mbosecke:
I'm surprised I never caught this. Thanks.

Issue #146, opened 2015-12-08
Add the ability to disable template cache during development phase


donnior:
Currently template cache is based on filename, if we already have one template page and update its content, the engine will use the old-cached template, unless we restart the whole engine.

We found this will be a big problem if we develop under servlet env. The spring-pebble integration solution  will always use the old content, so we need restart servlet container every time when we modified template's content.

For a production env, the default cache-enabled feature is OK, but for development env we hope can disable it.

Could you please think about it ? @mbosecke

ebussieres:
You don't need to modify anything in the code. In your config just set the template cache to null like this. Here's with the new 2.0 syntax but you can set the template cache to null with version 1.6.0 too.

```javascript
   public PebbleEngine pebbleEngine() {
        PebbleEngine.Builder builder = new PebbleEngine.Builder()
                .loader(this.templateLoader())
        if (this.isPebbleCacheInactive()) {
            builder.templateCache(null);
            builder.tagCache(null);
        }
        return builder.build();
```


The main advantage of having access to templateCache is that you can if you wish define your cache with the desired size (default is 200) or use a time to live cache for example.

But maybe just adding a cacheActive(false) method would be great. Something like

```javascript
    PebbleEngine.Builder builder = new PebbleEngine.Builder()
                .loader(this.templateLoader())
                .cacheActive(false);
```

If you define the template cache and set cacheActive to false, I think that it should disabled the cache.

```javascript
    PebbleEngine.Builder builder = new PebbleEngine.Builder()
                .loader(this.templateLoader())
                .templateCache(CacheBuilder.newBuilder().maximumSize(200).build())
                .cacheActive(false);
```

@mbosecke what do you think about it ?

donnior:
Agree with @ebussieres , add one cache-toggle method will make PebbleEngine more convenient for some config scenario, such as pebble-spring, we can config it like this: 

```
<bean id="pebbleEngine" class="com.mitchellbosecke.pebble.PebbleEngine">
    <property name="loader" ref="pebbleTemplateLoader" />
    <property name="cacheActive" value="false" />   //pass env variable, false for dev and true for prod environment
</bean>
``` 
we can easily use build script to pass different environment variables with same configuration code.

ebussieres:
I've just created a pull request for the cache active flag

mbosecke:
Implemented by #148

PR #147, opened 2015-12-08
support enable and disable template cache

donnior:
Use Guava's invalidate to disable template cache ( #146 ) , useful for constantly template content updating during development phase.

For example we can use it like this under SpringMVC

```
<bean id="pebbleEngine" class="com.mitchellbosecke.pebble.PebbleEngine">
    <property name="loader" ref="pebbleTemplateLoader" />
    <property name="enableTemplateCache" value="false" />   //false for dev and true for prod environment
</bean>
```

### DIFF
--- a/src/main/java/com/mitchellbosecke/pebble/PebbleEngine.java
+++ b/src/main/java/com/mitchellbosecke/pebble/PebbleEngine.java
@@ -67,6 +67,8 @@ public class PebbleEngine {
 
     private final ExtensionRegistry extensionRegistry;
 
+    private boolean enableTemplateCache = true;
+
     /**
      * Constructor for the Pebble Engine given an instantiated Loader. This
      * method does only load those userProvidedExtensions listed here.
@@ -116,6 +118,10 @@ public class PebbleEngine {
 
         try {
             final Object cacheKey = this.loader.createCacheKey(templateName);
+
+            if (!enableTemplateCache) {
+                templateCache.invalidate(cacheKey);
+            }
 
             result = templateCache.get(cacheKey, new Callable<PebbleTemplate>() {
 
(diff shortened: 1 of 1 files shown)


ebussieres:
You don't need to modify anything in the code. In your config just set the template cache to null like this. Here's with the new 2.0 syntax but you can set the template cache to null with version 1.6.0 too.

```javascript
   public PebbleEngine pebbleEngine() {
        PebbleEngine.Builder builder = new PebbleEngine.Builder()
                .loader(this.templateLoader())
        if (this.isPebbleCacheInactive()) {
            builder.templateCache(null);
            builder.tagCache(null);
        }
        return builder.build();
```javascript

**Comment by donnior** [comment]
@ebussieres Great，We are using 1.6.0 and just found the code to set template cache to null.
But I think expose one interface to enable or disable template cache is better than setting null, for example we can build project use same build script but with different environment variables.

I will close this PR and hope you can think about this.

**Comment by ebussieres** [comment]
The main advantage of having access to templateCache is that you can if you wish define your cache with the desired size (default is 200) or use a time to live cache for example.

But maybe just adding a cacheActive(false) method would be great. Something like

```javascript
    PebbleEngine.Builder builder = new PebbleEngine.Builder()
                .loader(this.templateLoader())
                .cacheActive(false);
```

If you define the template cache and set cacheActive to false, I think that it should disabled the cache.

```javascript
    PebbleEngine.Builder builder = new PebbleEngine.Builder()
                .loader(this.templateLoader())
                .templateCache(CacheBuilder.newBuilder().maximumSize(200).build())
                .cacheActive(false);
```

@mbosecke what do you think about it ?

ebussieres:
I think we should continue this conversation on the issue and close the PR

PR #149, opened 2015-12-16
Escape functions by default

samlanning:
This PR was motivated by our use internally at our company, and realising that functions are not escaped by default. Rather than explicitly making each function escape its output before returning, we thought it would be beneficial for functions to be escaped by default, and require explicit escaping.

To achieve this, the following changes have been made.

* Introduction of test cases to check the escape behaviour of basic functions and macros.
* The introduction of `RawString.java`: a class that the `EscapeFilter` ignores (as it is not a `String`).
* Functions and Macros are not now whitelisted. Instead macros return an instance of `RawString` when evaluated, and functions return whatever they would usually. This means that macros are still never escaped, and functions are by default, but can return an instance of `RawString` to prevent escaping. `parent()` and `block()` are not affected by this.
* Filter whitelists have been removed, and a filter can return raw output (just like functions) by returning a `RawString`.
* The core filters that handle `Strings` have all been updated to handle `RawStrings` too, and continue to fail with a class cast exception in any other case.

### DIFF
--- a/src/main/java/com/mitchellbosecke/pebble/PebbleEngine.java
+++ b/src/main/java/com/mitchellbosecke/pebble/PebbleEngine.java
@@ -408,17 +408,6 @@ public class PebbleEngine {
         }
 
         /**
-         * Marks a particular tag as safe to the built-in escaper extension.
-         *
-         * @param filter The name of the tag to be marked as safe
-         * @return This builder object
-         */
-        public Builder addEscaperSafeFilter(String filter) {
-            escaperExtension.addSafeFilter(filter);
-            return this;
-        }
-
-        /**
          * Adds an escaping strategy to the built-in escaper extension.
          *
          * @param name     The name of the escaping strategy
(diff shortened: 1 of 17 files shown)


mbosecke:
This seems reasonable to me. I am absolutely racking my own brain trying to think of why I decided to not escape functions in the first place. I even documented that they aren't escaped but for the life of me I can't think of why. 

It must have been an oversight but I'll be thinking about it a bit more before I merge this in. 

Thanks for the work you put into this.

samlanning:
Thanks for the friendly reception of this PR.

I look forward to hearing your decision.

samlanning:
I have some follow up commits to follow this which may need some more discussion: https://github.com/samlanning/pebble/tree/escape-functions-rawstring-tostring

It intends to make it a bit more obvious when you have mistakes using `RawStrings`.

samlanning:
Hey @mbosecke, any decisions made re: this PR?

mbosecke:
Thanks for the contribution, sorry it took so long to merge!

anasofiapaixao:
I was bumping pebble from 2.0.0 to 2.2.0 in one of our projects, and it appears that this pull request was merged somewhere in between. This has as a consequence that code using the methods removed stops compiling, effectively being a breaking change in a minor version bump. I can make a PR later, restoring the methods as @Deprecated no-ops and documenting that those calls became unnecessary; just figured you would want to know in the meantime.

anasofiapaixao:
I've made a pull request fixing the above: https://github.com/PebbleTemplates/pebble/pull/183

PR #150, opened 2015-12-17
Don't Just escape Strings

samlanning:
Escape anything, and not just `Strings`.

Previously, the `EscapeFilter` would ignore anything that wasn't a String, this
meant that an object being printed could contain some XSS content in its
`toString()` method (e.g standard java collections that contain a string with
XSS) that didn't get escaped, even if the escape filter was used explicitly.
The `EscapeFilter` now converts anything to a `String` with `toString()`, and then
proceeds as normal.

This PR will conflict with #149 and one or other of them will need to be rebased when the other gets merged, I will of course do this when necessary.

Also worth noting that even if there were no merge conflicts, tests would fail after both were to get merged in as `RawString` would no longer be ignored as required by `EscapeFilter` (so this change will also need to be made).

### DIFF
--- a/src/main/java/com/mitchellbosecke/pebble/extension/escaper/EscapeFilter.java
+++ b/src/main/java/com/mitchellbosecke/pebble/extension/escaper/EscapeFilter.java
@@ -77,10 +77,10 @@ public class EscapeFilter implements Filter {
     }
 
     public Object apply(Object inputObject, Map<String, Object> args) {
-        if (!(inputObject instanceof String)) {
+        if (inputObject == null || inputObject instanceof SafeString) {
             return inputObject;
         }
-        String input = (String) inputObject;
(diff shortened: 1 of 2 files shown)


mbosecke:
This one seems like a good idea but it appears to break a bunch of our existing tests.

samlanning:
Should we take a look at this one again do you think? What would be the best way to proceed?

karlpauls:
+1, I'd like to second this issue. It would be good to have toString results be part of the EscapeFilter checks.

As of the patch itself I have no idea if it make sense or not or what the problems are with it but it would be nice to have the functionality implemented :-)

samlanning:
Rebased and fixed issue (i was forgetting to not escape `SafeString`). All tests pass.

PR #152, opened 2015-12-17
Add PebbleEngine.Builder.build variant with custom base extensions

sgammon:
This fixes issue mbosecke/pebble#151, where default base extensions (i.e. `i18n`) can be substituted for user-provided ones, in case there are issues with the default set.

A new variant of `PebbleEngine.Builder.build` simply accepts a `List` of `Extensions`, which will be used along with any `userProvidedExtensions` otherwise given to `Builder`.

The existing API is unchanged, and returns the result of the new method, but with the existing list of default extensions.

### DIFF
--- a/src/main/java/com/mitchellbosecke/pebble/PebbleEngine.java
+++ b/src/main/java/com/mitchellbosecke/pebble/PebbleEngine.java
@@ -455,7 +455,6 @@ public class PebbleEngine {
          */
         public PebbleEngine build() {
             List<Extension> extensions = new ArrayList<>();
-            extensions.add(new CoreExtension());
             extensions.add(escaperExtension);
             extensions.add(new I18nExtension());
             return build(extensions);
@@ -466,9 +465,12 @@ public class PebbleEngine {
          *
          * @return A PebbleEngine object that can be used to create PebbleTemplate objects.
          */
-        public PebbleEngine build(List<Extension> extensions) {
+        public PebbleEngine build(List<Extension> baseExtensions) {
 
             // core userProvidedExtensions
+            List<Extension> extensions = new ArrayList<>();
+            extensions.add(new CoreExtension());
+            extensions.addAll(baseExtensions);
             extensions.addAll(this.userProvidedExtensions);
 
             // default loader
--- a/src/main/java/com/mitchellbosecke/pebble/PebbleEngine.java
+++ b/src/main/java/com/mitchellbosecke/pebble/PebbleEngine.java
@@ -440,6 +440,15 @@ public class PebbleEngine {
         }
 
         /**
+         * Return the active escaper extension, usually for overriding the list of base extensions.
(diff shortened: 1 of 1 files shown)


src/main/java/com/mitchellbosecke/pebble/PebbleEngine.java
```diff
@@ -440,17 +440,37 @@ public Builder cacheActive(boolean cacheActive) {
...
 
         /**
-         * Creates the PebbleEngine instance.
+         * Return the active escaper extension, usually for overriding the list of base extensions.
+         *
+         * @return The current escaper extension
+         */
+        public EscaperExtension getEscaperExtension() {
+            return escaperExtension;
+        }
```
sgammon: small getter for `escaperExtension`

src/main/java/com/mitchellbosecke/pebble/PebbleEngine.java
```diff
@@ -440,17 +440,37 @@ public Builder cacheActive(boolean cacheActive) {
...
+         * Creates the PebbleEngine instance, with default base extensions.
          *
          * @return A PebbleEngine object that can be used to create PebbleTemplate objects.
          */
         public PebbleEngine build() {
+            List<Extension> extensions = new ArrayList<>();
+            extensions.add(escaperExtension);
+            extensions.add(new I18nExtension());
+            return build(extensions);
+        }
```
sgammon: old method uses new method, no API changes required in non-GAE environments

src/main/java/com/mitchellbosecke/pebble/PebbleEngine.java
```diff
@@ -440,17 +440,37 @@ public Builder cacheActive(boolean cacheActive) {
...
+         */
+        public PebbleEngine build(List<Extension> baseExtensions) {
 
             // core userProvidedExtensions
             List<Extension> extensions = new ArrayList<>();
             extensions.add(new CoreExtension());
-            extensions.add(escaperExtension);
-            extensions.add(new I18nExtension());
+            extensions.addAll(baseExtensions);
             extensions.addAll(this.userProvidedExtensions);
```
sgammon: new method does the build work, and exposes a `baseExtensions` parameter

mbosecke:
I appreciate the effort but I'm not a fan of these changes. 

* It exposes the escaper extension which can cause concurrency issues. Removing this exposure was the reasoning I moved to the Builder pattern in the first place.
* It complicates the API. It's not obvious how to add a new extension vs replacing a core extension.
* Core extensions weren't ever intended on being replaced.

sgammon:
No worries, it was kind of a stab in the dark. How would you recommend going about a fix, if not via this route?

sgammon:
Ah, just saw your comment on the issue. I'll move over there.

PR #158, opened 2016-01-14
Fix some string concats outside of a StringBuilder

hectorlf:


hectorlf:
Oh, sorry, didn't expect that next commit would appear here. I'll revert it now.

mbosecke:
Thanks!

PR #161, opened 2016-01-19
Change ternary expression parsing to be more restrictive

hectorlf:
Just realized that this change may break existing templates :-/


mbosecke:
One of the biggest things I've learnt from this project is that "backwards compatibility" is a surprisingly ill-defined term. [Relevant xkcd comic.](https://xkcd.com/1172/).

hectorlf:
Hahaha, good one!

PR #177, opened 2016-02-29
Allow boolean arguments in method calls. Fixes #95.

### LINKED ISSUE #95: Calling Methods with non string parameter?
Afaik it is currently not possible to call a method that looks like this:
public void getAccounts(boolean removeZeroBalance)...
right? As a workaround I currently have to change the method parameter to String and parse it back to boolean inside the method. Is this the only way or is there a way to call methods that have non String parameter?

bjansen:


mbosecke:
Thanks! I appreciate it.

PR #179, opened 2016-03-02
Add filter for merging maps/lists/arrays

### LINKED ISSUE #164: Augment ConcatenateExpression to allow array, list and map parameters
I was adapting my templates when I found a use case where fusing two maps would've been handy. I remember there was something in the + operator, but it actually only adds Lists.

I'm not keen on overloading +, so what about adding the support to ~? Want a PR?

hectorlf:
Basic functionality is in place:
- maps with maps, maps with lists
- lists with lists, lists with maps
- arrays with arrays 

Fixes #164


hectorlf:
Shall I drop support for concatenating two lists from the + operator?

mbosecke:
This looks awesome, thanks a lot.

Regarding the "+" operator, I added some @Deprecated annotations and we'll consider removing the support for lists in the future but it can stay for now.

And I think we'll refrain from supporting non-collection objects with the merge filter for now as well. Keep it simple and if a good use-case comes up we can revisit the problem at that time.

PR #183, opened 2016-03-15
restore and deprecate addEscaperSafeFilter

anasofiapaixao:
This PR restores PebbleEngine Builder's addEscaperSafeFilter as a deprecated no-op, which was removed in a minor release (2.1.0). This allows delaying a breaking API change until a new major version is released.

### DIFF
--- a/src/main/java/com/mitchellbosecke/pebble/PebbleEngine.java
+++ b/src/main/java/com/mitchellbosecke/pebble/PebbleEngine.java
@@ -404,6 +404,18 @@ public class PebbleEngine {
         }
 
         /**
+         * Marks a particular tag as safe to the built-in escaper extension.
+         *
+         * @param filter The name of the tag to be marked as safe
+         * @return This builder object
+         * @deprecated tags are now escaped by default; this call should be removed
+         */
+        @Deprecated
+        public Builder addEscaperSafeFilter(String filter) {
+            return this;
+        }
+
+        /**
          * Adds an escaping strategy to the built-in escaper extension.
          *
          * @param name     The name of the escaping strategy


samlanning:
Although this *does* prevent a breaking API change, what it does mean is that if you put this noop method in, the behaviour is drastically different. And in fact can mask the fact that there is now a problem.

Any usage of this should instead update their filters to return a RawString, because unless they do that, their filters will be escaped when they shouldn't be, resulting in unexpected behaviour that may otherwise go unnoticed.

It makes sense to me to flag this up as something they need to update, rather than hide it and have a version of pebble that has potentially bad behaviour like this.

As such I am personally opposed to this change.

anasofiapaixao:
I see how removing the method may help some library users be aware of the change, as since it was released as minor already, a breaking change in a minor will raise more red flags in a dev's head than a mere deprecation would and lead someone to investigate further and find out what happened, just like I did. It is useful for people who may have both escaped and unescaped filters. However, the change is breaking only in the cases where functionality would be kept anyway; that is, you'll only see it if you have at least one filter that is escaped already, and **if you only had unescaped filters, their behaviour will be changed with zero warning**. I guess the nicest thing to do now that the minor's out of the bag already would be to purposely break the registration of extensions and give a heads-up about the change in javadoc? We all know not to fully trust semantic versioning is always respected, but there's bound to be an unlucky dev who will get caught by surprise for as long as there are people bumping from 2.0.x to 2.x.x, and it hurts both them and pebble.

In any case, I completely agree with you, that would be a whole other PR and something that ought to be done before even considering this one.

mbosecke:
Thanks anyways @anasofiapaixao  but I think I agree with @samlanning on this one. 

I made a judgement call to introduce a breaking API change (which may or may not have been a good decision on my part) but the damage is already done. I think trying to hide the issue at this point might make it even worse.

Issue #182, opened 2016-03-15
ConcurrentModificationException when importing macros in nested parallels blocks


karlpauls:
The fix for #159 just passes down the importedTemplates list to the embedded EvaluationContext. This fixes the problem that parallel blocks don't have access to macros imported into the outer context however, now we get a problem in the case where one is importing macros _inside_ the parallels block. This is concurrently modifying the list of importedTemplates (and technically, is leaking inner macros to the outer context).

karlpauls:
I can basically think of two easy ways to fix this. I created a pull request for each. Furthermore, they contain a test for this issue that does repeatable fail for me (depending on your hardware ymmv). 

The difference between the two is that one (the "snapshot") makes it so that the macros included in an inner context are not "leaked" to its outer context while the other (the "copyonwrite") does still "leak".  Personally, I think the "leaking" is not what i would expect so I'd go with the "snapshot" patch but I don't feel very strongly either way. 

Finally, the reason the test is writing in a bit of a roundabout way is that in case of a parallel tag, exceptions during the evaluation just bubble up to the Thread of the Executor (where they get swallowed) -- hence, from the outside (the template.eval) it looks like everything went ok while in reality a swallowed exception might have messed the result up. I think this is a bug in its own right which probably should be addressed in a separate issue/patch.

mbosecke:
Good catch, thanks for both the test and the proposed solutions. I went with the "snapshot" technique but I adjusted where the list is copied so that it only occurs when using the "parallel" tag.

PR #186, opened 2016-03-18
Fix ConcurrentModificationException when importing macros in nested parallels blocks (#182) by seeding the importedTemplates passed into the root EvaluationContext with a CopyOnWriteArrayList instead of an ArrayList.

karlpauls:
This patch fixes ConcurrentModificationException when importing macros in nested parallels blocks (#182) by seeding the importedTemplates passed into the root EvaluationContext with a CopyOnWriteArrayList instead of an ArrayList.

The benefit of seeding with a CopyOnWriteArrayList that keeps being passed down is that it potentially saves some memory however,  it does "leak" macros to the outer context in case of parallel blocks which is not what I would expect. Obviously, it fixes the ConcurrentModificationException as well. 

Furthermore, the patch includes a test for the issue that fails without the patch but passes with it on my system. Unfortunately, the test is done in a little roundabout way as the actual issue (the ConcurrentModificationException) is silently swallowed and doesn't fail the overall template execution (which I would consider a bug in its own right) -- hence, it just does run same same template one time without an Executor and one time with an Executor and compares the result (which isn't guaranteed to fail depending on you hardware - it does however, fail on my system everytime out of 20 runs).

### DIFF
--- a/src/main/java/com/mitchellbosecke/pebble/template/PebbleTemplateImpl.java
+++ b/src/main/java/com/mitchellbosecke/pebble/template/PebbleTemplateImpl.java
@@ -22,6 +22,7 @@ import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 /**
  * The actual implementation of a PebbleTemplate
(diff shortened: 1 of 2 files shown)


mbosecke:
I went with #185

PR #189, opened 2016-03-29
Improve handling of null values in boolean logic.

samlanning:
Something that has been frustrating me is the inconsistency in handling `null` values in if statements, and boolean expressions. Namely, `null` is implicitly `false` in `if` blocks, but when used in more complex expressions (using `not`, `and` or `or`), it is treated as an invalid operand. (This leads to iterative changes in templates potentially leading to errors where it makes no sense for that regression to happen).

This PR makes the following changes:
- null values are implicitly false when used with `not`, `or` and `and`.
- In strict mode, `null` values are not allowed in any of these operators, and also now not allowed in an `if` statement: previously, in strict mode this would render `no` rather than throwing an error:
  
  ``` django
  {% if null %}yes{% else %}no{% endif %}
  ```


mbosecke:
Sorry for the delay in responding. I'm a fan of the changes but it breaks backwards compatibility so it'll have to wait until the next major release.

mbosecke:
Merged onto the 3.0 branch

PR #194, opened 2016-04-15
Fixed NullPointerException in UnaryNotExpression

MichaelChampagne-Sepaq:
I had a case where the condition of a ternary expression was null. Now it's handled, test included.


MichaelChampagne-Sepaq:
I didn't see https://github.com/PebbleTemplates/pebble/pull/189 before submitting my pull request. I guess the only debate left is...
Since IF NULL is treated as false, what should IF NOT NULL do?

1. Be treated as the opposite (true) because of the NOT operator.
2. Still be treated as false, because the condition value is still not a boolean value.

ebussieres:
I vote for option 2

mbosecke:
The NPE has been fixed on the 3.0 branch via #189.

PR #210, opened 2016-07-21
Fix tests when run from a path containing spaces

toby-allsopp:
If the path containing the source code has spaces in it, then getResource
returns a URL with spaces replaced by %20, which, when treated as a file
name, fails to find anything.

The solution is to convert to a File object (via URI) as that process
takes care of decoding the %20 (and any other URL encoding that might
creep in).


mbosecke:
Thanks!

Issue #201, opened 2016-05-21
Impossible to reset a variable


C4Grey:
Variables inside templates cannot be reseted(although it is available in Twig), for example:
````code=twig
{% set total = 0 %}
{% for item in items %}
     {% set total = total + item.balance %}
{% endfor %}
Balance: {{ total }}
````
The "total" variable will be equal to 0 in the last line.
P.S.
It seems like a variable can be reseted outside a cycle, so maybe this issue is related to scoping.

**Comment by sunnypav** [comment]
I have also encountered the same issue which is working totally fine in Twig. Please fix.

**Comment by sunnypav** [comment]
I am also facing this issue which is totally working fine in Twig. Please fix.

**Comment by stanislas-m** [comment]
I can't reproduce this problem with this test:

```java
@Test
public void testReSetInForLoop() throws PebbleException, IOException {
    PebbleEngine pebble = new PebbleEngine.Builder().loader(new StringLoader()).strictVariables(false).build();

    String source = "{% set total = 0 %}{% for item in items %}{% set total = total + item.balance %}{% endfor %}{{ total }}";
    PebbleTemplate template = pebble.getTemplate(source);
    Map<String, Object> context = new HashMap<>();
    List<Map<String,Object>> items = new ArrayList<>();
    for (int i = 1; i < 4 ; ++i) {
        Map<String, Object> item = new HashMap<>();
        item.put("balance", i);
        items.add(item);
    }
    context.put("items", items);

    Writer writer = new StringWriter();
    template.evaluate(writer, context);
    assertEquals("6", writer.toString());
}
```

Do you have more details about the context?

C4Grey:
I am sorry for late response, moved to another project for a while...well, here are some details:
- Spring Boot 1.4.0.M1
- Pebble added via the com.mitchellbosecke:pebble-spring-boot-starter:2.2.1 dependency
- com.mitchellbosecke.pebble.loader.FileLoader used for templates loading
- Caching disabled

toby-allsopp:
Try nesting inside another for loop. This code in ForNode.java is responsible for this behaviour, I think:

```
            /*
             * Only if there is a variable name conflict between one of the
             * variables added by the for loop construct and an existing
             * variable do we push another scope, otherwise we reuse the current
             * scope for performance purposes.
             */
            if (scopeChain.currentScopeContainsVariable("loop") || scopeChain
                    .currentScopeContainsVariable(variableName)) {
                scopeChain.pushScope();
[...]

PR #211, opened 2016-07-21
Allow the set tag to update existing values in parent scopes

toby-allsopp:
This fixes a problem when using the set tag in a nested for tag. The for
tag pushes a new scope if there is already a "loop" variable in scope. If
a set tag executes in that nested scope, the changes are undone at the end
of the for tag, which is not what users expect.

Example:

{% set x = 1 %}
{% for i in 1..1 %}
  {% for j in 1..1 %}
    {% set x = 2 %}
  {% endfor %}
{% endfor %}
{{ x }}

Previously this would yield "1"; now it will yield "2".

### DIFF
--- a/src/main/java/com/mitchellbosecke/pebble/node/SetNode.java
+++ b/src/main/java/com/mitchellbosecke/pebble/node/SetNode.java
@@ -30,7 +30,7 @@ public class SetNode extends AbstractRenderableNode {
 
     @Override
     public void render(PebbleTemplateImpl self, Writer writer, EvaluationContext context) throws PebbleException {
-        context.getScopeChain().put(name, value.evaluate(self, context));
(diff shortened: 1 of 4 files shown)


toby-allsopp:
This is a fix for issue #201, in which set tags inside nested for loops are undone at the end of the for loop.

mbosecke:
This is a clever solution, thanks. I'll have to think a bit more about scenarios where a user may have been relying on the old behavior but I'm optimistic that this is the way to go.

toby-allsopp:
Cool, thanks for taking a look.

FYI, we're no longer relying on this change. We were using it so we could implement a global counter that increments throughout the template, but we discovered that that approach can't work through macros so I changed to a static variable in a custom tag, which is totally thread-unsafe but works fine for our use case.

PR #212, opened 2016-07-26
Add tests for dynamic include and import

flof:
This commit adds tests for the feature that one can use a dynamic expression to determine the template in the include- and import-tag.


mbosecke:
Great! I'm surprised we didn't have this already.

PR #215, opened 2016-08-02
Stop ScopeChain lookups when first scope is local

toby-allsopp:
A local scope should stop the lookup from proceeding along the chain. This
was working in all cases except when the very first scope in the chain was
local.


toby-allsopp:
I only just realised that this has a rather important effect on how default parameters to macros are evaluated. Without this pull request, they are able to see variables that are in the caller's scope and with this pull request they are not.

Which is the correct behaviour?

toby-allsopp:
Hi, thanks for the merge but did you see my comment from Aug 22? This change breaks backward compatibility - we reverted it from our internal fork because we had people relying on the old behaviour.

PR #219, opened 2016-08-05
Add hasBlock method to PebbleTemplateImpl

toby-allsopp:
This is useful in certain NodeVisitors.

### DIFF
--- a/src/main/java/com/mitchellbosecke/pebble/template/PebbleTemplateImpl.java
+++ b/src/main/java/com/mitchellbosecke/pebble/template/PebbleTemplateImpl.java
@@ -192,6 +192,16 @@ public class PebbleTemplateImpl implements PebbleTemplate {
     }
 
     /**
+     * Checks if a block exists
+     *
+     * @param blockName The name of the block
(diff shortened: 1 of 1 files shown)


mbosecke:
Can you provide an example of a node visitor that would benefit from this addition?

toby-allsopp:
In my NodeVisitor, I want to only visit the version of a block that will actually be used, i.e. if a child template overrides a block I want to avoid visiting the parent's version, so I have this method:

```java
    private Deque<PebbleTemplateImpl> m_children;

        @Override
        public void visit(BlockNode node)
        {
            for (PebbleTemplateImpl child : m_children)
            {
                if (child.hasBlock(node.getName()))
                {
                    return;
                }
            }
            node.getBody().accept(this);
        }

        @Override
        public void visit(ExtendsNode node)
        {
            final Expression<?> parentExpression = node.getParentExpression();
            parentExpression.accept(this);
            m_children.push(getTemplate());
            loadAndVisitTemplate(parentExpression);
            m_children.pop();
        }
```

PR #225, opened 2016-09-04
Fixes #217: Fix regression of broken SafeString concatenation

### LINKED ISSUE #217: Can not concatenate SafeString
In Pebble 1.x I used to be able to concatenate the output of a macro with the output of another macro.  Something like:

```twig
{{ macro1() + macro2() }}
```

In Pebble 2.x I can no longer do that and I think the problem is [this](https://github.com/PebbleTemplates/pebble/blob/master/src/main/java/com/mitchellbosecke/pebble/utils/OperatorUtils.java#L39-L46) hunk of code which only checks for `String` and does not consider `SafeString`.

gitblit:
This fix is working for me, however you may have an alternate implementation suggestion.

### DIFF
--- a/src/main/java/com/mitchellbosecke/pebble/utils/OperatorUtils.java
+++ b/src/main/java/com/mitchellbosecke/pebble/utils/OperatorUtils.java
@@ -8,6 +8,8 @@
  ******************************************************************************/
 package com.mitchellbosecke.pebble.utils;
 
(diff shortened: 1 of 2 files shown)


gitblit:
Any thoughts? I'd like to update to the 2.x Pebble releases but this bug is blocking me.

gitblit:
Thanks for the merge.  Would you be willing to deploy a point release with this fix?

Issue #125, opened 2015-11-15
No newline after expression that is at end of the line


zylom:
When an expression is at the end of a line, there seems to be no newline character to seperate that line with the next line.

``` javascript
<html>
  <body>
    <div id="content">
      {{ mainContent }}
    </div>
  </body>
</html>
```

In the previous example, given that the `mainContent` variable will be replaced with `"Hello, world"`, the output will be:

``` html
<html>
  <body>
    <div id="content">
      Hello, world    </div>
  </body>
</html>
```

As you can see, there is no newline after `"Hello, world"`.  The `</div>` closing tag with its four spaces of indentation should be on the next line, not on the same line that `"Hello, world"` is on.

mbosecke:
This is a fairly common complaint but the original intention of removing the newline after every pebble expression is because otherwise you would have a lot of blank lines. For example:

    <p>
    {%  if true %}
        lorem ipsum
    {% endif %}
    </p>

would print with two blank lines:
   
    <p>
    
        lorem ipsum
    
    </p>

I'm not necessarily opposed to change but we will want to make sure we maintain the above requirement.

r0b0tbuilder:
To comply with that requirement, I was thinking of implementing this enhancement as follows:

Newline characters are removed only if that newline is after an expression, otherwise (such as in the case when it is at the end of a control-flow delimiter) the newline will be preserved.

If that sounds good, I'll start working on a pull request. :+1:

mbosecke:
Let me just reword what you're saying to make sure we're both on the same page:

If the newline immediately follows a {% ... %} statement then it will be removed but if it follows a {{ ... }} statement it will be preserved. Correct?

I would agree that this is a step in the right direction and I would probably accept a pull request for this, although be forewarned that it may not be an easy task (although it's been awhile since I looked at the relevant code). 

I do also want to mention that we can optionally take this even a step further. Ideally, the newline would be preserved in the following example too even though it follows a {% ... %} statement:

```
Subject: {% if true %} subject line {% endif %}
Message: message
```

I think the ideal rule would be to only remove newlines if that entire line of text contains nothing but a single {% ... %} statement (and whitespace).

r0b0tbuilder:
Yes, that is what I was trying to convey. We're both on the same page :smiley: . I agree with your suggestion, and I will try my best to implement this as you suggested.

r0b0tbuilder:
[...]

PR #226, opened 2016-09-22
Issue #125 - Disable trimming of the first newline after a tag.

electrotype:
Adds the possibility to use ".enableNewLineTrimming(false)" when  building a PebbleEngine instance so the first newline following a Pebble tag is not trimmed. Trimming the first newline stays the default behavior.


electrotype:
Maybe the code should be changed or a comment should be added to specifiy that if  a custom `Syntax` object is provided when building the `PebbleEngine` instance, then the `enableNewLineTrimming(...)` has no effect.

PR #227, opened 2016-09-26
Added Obfuscate Filter

rygel:
I created an obfuscate filter to enable the obfuscation of email addresses. I have taken the obfuscation source code from pegdown (https://github.com/sirthias/pegdown/blob/master/src/main/java/org/pegdown/FastEncoder.java).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+- Added "obfuscate" filter
+
 ## v2.2.2 (2016-07-03)
 - Improved error when attempting to iterate over a non-iterable object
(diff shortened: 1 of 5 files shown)


ebussieres:
I don't think it should be included in the core library. I don't see a lot of use case where it is necessary. I think it's the perfect use case for a custom filter in your app, but not include it in the core lib.

What does everyone else think about it ?

nikku:
Could be an extension (i.e. pebble-contrib) project?

mbosecke:
I have to agree with @ebussieres that this doesn't belong in the core library. Pebble is a general-purpose template engine used for contexts other than just HTML/web and this is feature is really only valuable for web applications.

PR #236, opened 2016-10-26
fixing tests on windows

menostos:
tests were failing due to different line endings coming from System.lineSeparator() and those which are located in test template files. Hope this doesn't breaks other systems.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,4 @@
+root = true
+
+[*]
+end_of_line = lf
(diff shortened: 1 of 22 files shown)


menostos:
I also added some functionality to configure the CoreExtension. It gives the possibility to disable it completly or just some elements.

mbosecke:
I don't get any errors when I run the tests on my windows machine. Try setting the git configuration `core.autocrlf` to `true` and see if that helps. That setting will convert all line endings to `\r\n` when you checkout the project and therefore the `System.lineSeparator()` invocation within the tests should be correct.

Also, I don't plan on officially supporting the removal of the core extension. There is code throughout the library that assumes the core extension is enabled and removing it can cause some undefined results. If a user really wants to change the core they'll have to fork the project and remove it themselves.

PR #239, opened 2016-11-14
Replacing Guava cache with the one from Caffeine

robertotru:
I made the replacement of the cache, updated some dependencies' version and fixed the involved unit tests.

Instead of creating a cache with max size set to zero, I decided to set the cache to `null`. I believe it should be faster to not pass through the cache, since it would always creates a `CompletableFuture` to compute the value that we have to return.


mbosecke:
Great contribution; it's very much appreciated. Thanks for cleaning/updating a bunch of tests as well.

robertotru:
@mbosecke don't mention it. I use Pebble a lot in my private projects, it was a pleasure.
By the way, it would be probably worth to make the caching a little bit more customizable. It could be useful to let the use specify the cache size, as well as introducing tags for time based cache. The first is an easy-to-do change. The latter may have some implications, I actually have no clear opinion on who should set the age for a template (i.e. is the view or the controller the owner of this piece of information?).

Issue #235, opened 2016-10-19
Feature Request: String Interpolation


DanStout:
Twig offers a feature called [string interpolation](http://twig.sensiolabs.org/doc/templates.html#string-interpolation) which they describe like so:

> String interpolation (#{expression}) allows any valid expression to appear within a double-quoted string. The result of evaluating that expression is inserted into the string:

Example:

```
{{ "Hello #{name}, you have #{msgs} messages, #{msgs - read} unread" }}
```

Compared to the current form:

```
{{ "Hello "+ name + ", you have "+ msgs + "messages, " + (msgs - read) + "unread" }}
```

I was planning to implement this in a filter that would work like:

```
{{ "hello #{name}" | interp }}
```

Where the input would be parsed, and any expressions would be evaluated using a separate PebbleEngine using a StringLoader. I didn't see a convenient way of accessing evaluating the string template with the correct context (I was planning to get it via EvaluationContext "_context" argument to the filter, but that didn't really work out). 

I feel like this might be better implemented as part of the engine core rather than an extension both for ease of implementation and because having to add " | interp" after every interpolated string kind of cancels out the reason to use it in the first place.

mbosecke:
I would definitely be on board with this being a core feature, implemented similarly to how Twig has done it. Pull requests are welcome.

DanStout:
OK. I started working on it, and will make a pull request in the near future.

Issue #230, opened 2016-10-07
A way to evaluate a String inside a template


electrotype:
Let's say using this Map :

```
user
 |___ firstName : "John"
 |___ lastName : "Doe"
```

Then something like this would be nice:

```
{% evaluate firstName = "user.firstName " %} 
```

==> 'John'

I'm not sure if this would be useful for many people, but it would be for me!

ebussieres:
Why don't you just use something like this 

```
{% set firstname = user.firstName %}
```

electrotype:
@ebussieres I agree there is a critical point missing in my question! : The "user.firstName" key _is dynamically built_, I don't know it at compile time!

This would be a better example :

    {% set key= user + ".firstName " %} 
    {% evaluate result = key %}

ebussieres:
Ah ok i see now ! It don't think it's supported but adding a evaluate tag would be useful I agree.

mbosecke:
If it's only the attribute name that is dynamic but the root variable isn't, you can use the square bracket notation to look up dynamic attributes: 

```
{% set attribute = 'firstName' %}
{{ user[attribute] }} 
```

But if the root variable is dynamic, i.e. "user", then I don't think there's currently a good solution. Feature request #235 is for string interpolation which I believe will solve your issue. If you think that's the case, we'll close this task as a duplicate.

electrotype:
If it's possible to pass a root variable name to that string interpolation construct, I'm pretty sure it would do the trick, yes! No problem for me if you close this task as a duplicate. Thanks.

electrotype:
So, after checking the PR #240 , the current suggested String Interpolation solution doesn't help with this issue.

mbosecke:
Yup, I was thinking about this issue a bit more and I think we'll need to come up with something a bit more creative. I'll have to take a look at what other template engines are doing. I believe twig uses a special  variable called `_context` that can be used to access dynamic variables, ex. `{{ _context[variableName] }}` but I'm not a fan of it.

DanStout:
This is pretty hacky, but if this is a somewhat special case, couldn't you implement a function like this:

```
private static class Func implements Function {
	@Override
	public List<String> getArgumentNames() {
		return Arrays.asList("name");
	}

	@Override
	public Object execute(Map<String, Object> args) {
		String key = (String) args.get("name");

		EvaluationContext ctx = (EvaluationContext) args.get("_context");
		ScopeChain chain = ctx.getScopeChain();

[...]

PR #240, opened 2016-11-14
Add String Interpolation (as per issue #235)

DanStout:
I also made ConcatenateExpressions be counted as safe when they are just made up of two LiteralStringExpressions, and also added a few toString implementations while I was debugging.

Also, I forgot to turn off Eclipse's format-on-save actions, so some of the whitespace & minor formatting changed a bit. Sorry about that!


electrotype:
Nice PR!!

I though this `String Interpolation` thing would help with this [issue](https://github.com/PebbleTemplates/pebble/issues/230) but, if I understand correctly, this kind of code would not work right? :

    @DanStout 
    public void testDynamicVariableInterpolation() throws PebbleException {

        String source = "{% set key = 'name' %}" +
                        "{{ \"Hello, #{key}\" }}";

        Map<String, Object> ctx = new HashMap<>();
        ctx.put("name", "joe");

        assertEquals("Hello, joe", evaluate(source, ctx));
    }

Any way you see to be able to evaluate a variable by dynamically generate its name? Maybe it's already possible in your solution and I simply don't see how?

DanStout:
Hmm.. I don't think so, or at least, not in its current form.

This string interpolation works exactly like regular parsing. The parsing occurs inside `ExpressionParser` like so:

```
stream.expect(Token.Type.STRING_INTERPOLATION_START);
nodes.add(parseExpression());
stream.expect(Token.Type.STRING_INTERPOLATION_END);
```
And there is nothing in `parseExpression()` which acts differently when you are in the string interpolation state.

PR #241, opened 2016-11-16
DynamicAttributeProvider

electrotype:
This is a feature I suggest and that would be really useful for me. It consists in a `DynamicAttributeProvider` interface which, when implemented by an object, tells the expression parser that this object is able to provide attributes _dynamically_, given their names and the potential arguments.

For example :

    public class MyObject implements DynamicAttributeProvider {

        @Override
        public boolean canProvideDynamicAttribute(Object attributeName) {
            
            if("firstname".equals(attributeName)) {
                return true;
            }
            return false;
        }

        @Override
        public Object getDynamicAttribute(Object attributeName, Object[] argumentValues) {
            
            if("firstname".equals(attributeName)) {
                return "John";
            }
            return null;
        }

        public String getSurname() {
            return "Doe";
        }
    }

Then `"{{ object.firstname }} {{ object.surname }}"` => **John Doe**

-------------------------

So, prior to check for :

- `getXXX `method
- `isXXX `method
- `hasXXX `method
- `XXX `method
- `XXX `field

the parser will call `DynamicAttributeProvider#canProvideDynamicAttribute(attributeName)` to see if the current object is able (and wants) to generate the attribute by itself. If so, the `DynamicAttributeProvider#getDynamicAttribute(attributeName, argumentValues)` method will be called.

This feature will help me since I do have a solution for issue #230 , but it involves adding a duplicate of all the `Map<String, Object>` parameters for the template to a special object and then add this object itself to the `Map`. With this feature, I wouln't have to duplicate anything.

Those changes shouldn't impact any existing code.

### DIFF
--- a/src/main/java/com/mitchellbosecke/pebble/extension/DynamicAttributeProvider.java
+++ b/src/main/java/com/mitchellbosecke/pebble/extension/DynamicAttributeProvider.java
@@ -14,8 +14,7 @@ public interface DynamicAttributeProvider {
     public boolean canProvideDynamicAttribute(Object attributeName);
     
     /**
-     * Returns the attribute associated to the specified key
-     * or <code>null</code> if not found.
+     * Returns the attribute given the specified name and arguments.
      */
     public Object getDynamicAttribute(Object attributeName, Object[] argumentValues);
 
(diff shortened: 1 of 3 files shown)


ebussieres:
@mbosecke I just committed some modifications to this PR. Do you feel comfortable with this addition for the problem discussed in issue #230 ? I think

PR #251, opened 2017-02-03
implemented Java8 Date API support on date filter

alcmaeo:
added LocalDate, LocalTime, LocalDateTime and ZonedDateTime support for date filter. added fallback to ISO_DATE_TIME pattern, if no pattern was defined. an exception will be thrown, if the pattern can not applied to date object. example LocalTime object with a time pattern.


ebussieres:
Thanks !

PR #253, opened 2017-02-07
Macros are called too often, if in an include hierarchy

jochen777:
I just discovered, that macros are called too often if you frequently import the same macro in different templates in a hierarchy.

Example:
"index" (import and call macro A, include template "child") -> (first call of macro A)
"child" (import and call macro A) -> (second call of macro A)

In fact, the macro A is currently called 3 times! Once in the "index" template and 2x in the "child" template, because it searches in the hierarchy and executes each macro in the upper hierarchy.

This fix solves this issue (hopefully). I added a test to show the issue.

------------
disclaimer: I really like pebble and we use it in a large site. Thank you for your work!


ebussieres:
Great work thanks !

PR #257, opened 2017-03-14
Removing break; from scope checking

wildunne:
This particular line of code is causing issues - while rendering the
template we are finding that the second item in the stack of 6 is
marked as `isLocal`, and this is preventing the template from 
finding the item that is in the fourth item of the stack.

Maybe I am missing the point of this line of code - but it is
certainly causing an issue.


wildunne:
The intention of this line of code is to limit the accessibility 
of variables to their scope - this is my mistake. Happens due
to trying to access a variable from inside a function

PR #265, opened 2017-05-02
getRawTemplate method in PebbleEngine, fixed info about line number when macro is invalid

jajcek:


ebussieres:
I just made a commit to output filename and line number which is way much simplier. You can have a look [here ](https://github.com/PebbleTemplates/pebble/commit/15ab669896909ca497e3672f1088dca767a82697)

Thanks a lot for your contribution, much appreciated.

jajcek:
I totally agree your solution is much simpler, but I have two questions:

1. Is there any possibility to get template as a string without ugly casting the loader (this is why I have added getRawTemplate)?

```
DelegatingLoaderCacheKey cacheKey = ((Loader<DelegatingLoaderCacheKey>) engine.getLoader())
                .createCacheKey("D:\\aaa.txt");
Reader reader = ((Loader<DelegatingLoaderCacheKey>)engine.getLoader()).getReader(cacheKey);
```

2. When do you plan to release 2.4.0?

ebussieres:
1. You can try this

```
Loader loader = engine.getLoader();
Object cacheKey = loader.createCacheKey("template.peb");
Reader reader = loader.getReader(cacheKey);
```

2. I can't do the release. Only Mitchell can do it. But I expect it to be really soon

PR #269, opened 2017-05-22
Extend PebbleTemplate to allow rendering of single blocks

tx-francesco-figari:
I have recently adopted Pebble for rendering html views in an enterprise application, and thought it could be useful for emails as well. It seems sensible to describe subject, text and html body in the same template, but this would require rendering them as separate blocks.

This PR extends `PebbleTemplate` and `PebbleTemplateImpl` with a new `evaluateBlock()` method that allows rendering of single blocks, similar to the Twig [renderBlock()](https://twig.sensiolabs.org/doc/2.x/api.html#rendering-templates) method.

Happy to improve this if I have missed an easier way!


tx-francesco-figari:
Thank you for the fixes!

PR #270, opened 2017-05-23
use templateName to load Reader, instead of cacheKey

yanxiyue:
consider this case,  some app has multi templates package for different sub domain, each package has same sub directory structure and same template name.  developer can implement Loader to  switch templates package. if enable cache,  there will be a problem， same templateName -> multi cacheKey

so, I think, 
cacheKey should be used to mapping template instance only. it should not be used to load reader.
 If necessary,  the loader which can invoke the createCacheKey method within Loader#getReader.

### DIFF
--- a/src/main/java/com/mitchellbosecke/pebble/PebbleEngine.java
+++ b/src/main/java/com/mitchellbosecke/pebble/PebbleEngine.java
@@ -123,7 +123,7 @@ public class PebbleEngine {
 
                     LexerImpl lexer = new LexerImpl(syntax, extensionRegistry.getUnaryOperators().values(),
                             extensionRegistry.getBinaryOperators().values());
-                    Reader templateReader = self.retrieveReaderFromLoader(self.loader, cacheKey);
+                    Reader templateReader = self.retrieveReaderFromLoader(self.loader, templateName);
                     TokenStream tokenStream = lexer.tokenize(templateReader, templateName);
 
                     Parser parser = new ParserImpl(extensionRegistry.getUnaryOperators(),
@@ -161,16 +161,12 @@ public class PebbleEngine {
      * to handle the generic cast.
      *
      * @param loader   the loader to use fetch the reader.
-     * @param cacheKey the cache key to use.
+     * @param templateName the template name to use.
      * @return the reader object.
      * @throws LoaderException thrown when the template could not be loaded.
      */
-    private <T> Reader retrieveReaderFromLoader(Loader<T> loader, Object cacheKey) throws LoaderException {
-        // We make sure within getTemplate() that we use only the same key for
-        // the same loader and hence we can be sure that the cast is safe.
-        @SuppressWarnings("unchecked")
-        T casted = (T) cacheKey;
-        return loader.getReader(casted);
+    private <T> Reader retrieveReaderFromLoader(Loader<T> loader, String templateName) throws LoaderException {
+        return loader.getReader(templateName);
     }
 
     /**
--- a/src/main/java/com/mitchellbosecke/pebble/loader/DelegatingLoader.java
+++ b/src/main/java/com/mitchellbosecke/pebble/loader/DelegatingLoader.java
@@ -53,16 +53,15 @@ public class DelegatingLoader implements Loader<DelegatingLoaderCacheKey> {
 
 
     @Override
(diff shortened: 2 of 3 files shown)


yanxiyue:
sorry for my poor english

cjbrooks12:
As it appears this repo is dying, I have decied to create and maintain a new fork of Pebble at [JavaEden/Pebble](https://github.com/JavaEden/pebble), and I would love to get this PR merged into the new repo. Please take this PR and open a new one to the `dev` branch of `JavaEden/pebble`.

The future of Pebble has been shaky for a while now, but I refuse to let Pebble die. It is a great project, and I hope to give it same love and care that the original maintainers have as they built Pebble. If you have any questions about the future of Pebble, please reach out on Gitter. Let's keep this community alive, I hope to see you over at the new location!

**New Pebble Home**
- Original announcement of new ownership: https://github.com/PebbleTemplates/pebble/issues/311#issuecomment-379496941
- https://github.com/JavaEden/pebble
- https://bintray.com/javaeden/Eden/Pebble
- https://gitter.im/JavaEden/pebble

ebussieres:
I don't really understand your use case here. Why does a app would have multiple packages with same sub directory structure and same template name ? Can you elaborate a little ?

yanxiyue:
Sorry, the use case description is not very clear.

For example, a site may have different front-ends such as PC, WAP, H5 and so on. Different front-end may have the same page architecture, but render different content. Multi template packages can be used to implement free switching of different front-ends.

But this use case is not the key point.  **The key point is that `cacheKey` should only be used to find the template in the cache,  should not be used to load template**

ebussieres:
Well I see your point and I don't understand why DelegatingLoaderCacheKey exists at first. We should probably use template name for loading template as well as the cache key. DelegatingLoaderCacheKey basically consists of a list of the same template name.

I'll check that and probably drop DelegatingLoaderCacheKey

ebussieres:
At first, loading template used only template name. The change was introduced with 8bcaf4e262c09d73f043c5d5cbfd881dcf417c05

thomashunziker:
We use Pebble to generate for multiple users HTML output. However the source may be called the same. Example:
User 1: has for the template name 'xyz.twig' something like '{{ test.output }}'
User 2: has for the template name 'xyz.twig' something like 'Other {{ test.output }}'.

By applying the delegation pattern here you can have a different caching strategy which considers from the context if it was user 1 or user 2. If you reduce this to a String you cannot do such a thing.

yanxiyue:
@thomashunziker 
Maybe I don't particularly understand your case.
Let me try to describe your usage, 
that is to say that you implement a custom loader, and then use `DelegatingLoaderCacheKey` to save some extra context data, right?

I guess that you save extra context data (which get from ThreadLocal) into `DelegatingLoaderCacheKey`? 
If so, can you directly get these data again in `getReader` method?

I'm worried that I misunderstood. If you can simplify the code and post it up, it will be easy for me to understand.

thomashunziker:
@yanxiyue 
I understand what you are aiming for: The ThreadLocal variable can be accessed also in the reader since this should be the same thread. As long as you are not using any parallel execution etc. From my perspective the parallel execution feature is not an important one because the gains are only limited. 

There is one not so obvious drawback by using the ThreadLocal within the reader: The cache has to be aware of the different cache key strategy. Means you cannot reuse an existing cache without wrapping one and rewrite it. So the cache key and the cache storage are not anymore independent if you switch to a simple string based solution.

yanxiyue:
@thomashunziker 
the **createCacheKey** method will be  retained, and different cache key strategy still can be implemented  by this method.

ebussieres:
@thomashunziker Please check my last comment

hectorlf:
Sadly, that wouldn't work because the template name is just a piece of the puzzle. If I understand Thomas correctly, the loader augments the template name to form the final template name (something like "user123:templateXyz.pebble"). Given this, the only object that knows how to find the cached instance is the Loader.

But my question now is, wouldn't generating the full template name in app code instead of the loader work?

Friendly reminder anyway: if this breaks Thomas's set up, we need to move this to 3.0. God knows who else could be depending on that functionality by now.

thomashunziker:
I suggest that you adjust it as you need. For it is not a big deal we currently maintain anyway a separate version. If we are going to update to the latest Pebble version we may start issuing again merge requests. From my understanding it should be possible to use a string as a key as long as the concurrent features are not used.

Issue #293, opened 2017-09-20
performance degradation


vlalykin:
Hi.
The last two versions consistently degraded performance by 10% each. Thus, version 2.4 compared to version 2.2 has a -20% performance.
The measurements were performed using the project [mbosecke/template-benchmark](https://github.com/mbosecke/template-benchmark).
![results](https://user-images.githubusercontent.com/5020203/30638153-1f435784-9e03-11e7-8f7d-720083368a6e.png)

slandelle:
One of the culprits seems to be 035106f4a5411a52ccc896ceb5af43342548b18b.

The `object instanceof DynamicAttributeProvider` branch was moved in front so the `instanceof`check is performed on every `GetAttributeExpression#evaluate` call. It seems this is pretty expensive, probably because `DynamicAttributeProvider` is an interface so the JVM has to scan the whole type hierarchy. It seems the JIT  doesn't get rid of this branch, even though there's no class implementing `DynamicAttributeProvider` loaded :(

```
java version "1.8.0_144"
Java(TM) SE Runtime Environment (build 1.8.0_144-b01)
Java HotSpot(TM) 64-Bit Server VM (build 25.144-b01, mixed mode)
```

Current 2.4.1-SNAPSHOT master:
```
Benchmark          Mode  Cnt      Score     Error  Units
Pebble.benchmark  thrpt   50  24099,151 ± 380,110  ops/s
```

With `object instanceof DynamicAttributeProvider` commented out (not familiar enough with the code to figure out where it should be moved):
```
Benchmark          Mode  Cnt      Score     Error  Units
Pebble.benchmark  thrpt   50  26829,292 ± 437,474  ops/s
```

I also applied the change on 035106f4a5411a52ccc896ceb5af43342548b18b and validated it restored performance compared to previous commit.

slandelle:
The other culprit seems to be the fix for #165 (shipped in 2.2.1) that introduced `MemberCacheKey`. I guess it causes extra allocation and more complex HashMap lookup.
I guess there's nothing that can be done about it :(

slandelle:
Hi @ebussieres 

I have 2 commits that partially restore previous perf:

* https://github.com/gatling/pebble/commit/3c63f078cee2ce79262dd92f429333d5043ad25e
* https://github.com/gatling/pebble/commit/fa410d9b0d210475e57a10544f683b178e772101

Would you be interested in cherry-picking them?

PR #326, opened 2018-04-08
Handle DynamicAttributeProvider with a lower priority than Maps and Arrays, see #293

slandelle:
Motivation:

#241 introduced `DynamicAttributeProvider`. Type checking in
`GetAttributeExpression` seems to be pretty expensive, so as
`DynamicAttributeProvider` is not commonly used, Maps and Arrays
branches should have higher priority.

Modification:

Move the `DynamicAttributeProvider` branch down.

Result:

+10% throughput


slandelle:
@ebussieres PTAL

slandelle:
@ebussieres Should new PR be done against master of 3.0 branch?

ebussieres:
Master plz

PR #328, opened 2018-04-08
Simplify CacheKey hierarchy

slandelle:
Part of the performance regression observed in #293 is caused by the
new cache keys.

Introducing a new key object has of course a performance cost, but the
current implementation can be improved.

Modifications:

* Drop useless BaseTagCacheKey abstraction
* Make CacheKey final and static (not bound to CacheNode instance
context).

Result:

Clean code. +2% perf


ebussieres:
Thanks a lot

PR #329, opened 2018-04-13
restricting an access to class/getClass attribute

jajcek:
added restriction so that you can't access class/getClass attribute from an object, e.g.: `"{{ object.class }}"`. By having an access to it you can execute a shell command.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - Fix performance regression issues (#326, #328)
 - Add Reverse filter (#315) 
 - Make "loop.length" and "loop.revindex" be lazy evaluated (#279)
+- Fixed security issue which allowed to execute shell command (by having access to Java's Class object)
 
 ## v2.4.0 (2017-06-04)
 - Add arrays support for iterable test (#254)
--- a/src/main/java/com/mitchellbosecke/pebble/error/ClassAccessException.java
+++ b/src/main/java/com/mitchellbosecke/pebble/error/ClassAccessException.java
@@ -1,0 +1,18 @@
+/*******************************************************************************
+ * This file is part of Pebble.
+ *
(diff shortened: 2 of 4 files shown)


cjbrooks12:
This would be a breaking change, as I have templates that depend on getting the class instance from the template. In addition, I do not believe this really solves the issue, as one could write a filter to return the Class, or provide another method as a workaround. The only way to fully prevent shell execution is to install a SecurityManager.

cjbrooks12:
One alternative might be to allow access to getting a Class, but in the GetAttributeExpression, fail-fast if the object is an instance of `java.lang.Class`.

jajcek:
Hi @cjbrooks12,

thank you for being interested in the topic. However, I am not sure if I understand everything correctly. The fix is supposed to prevent people trying to access class object only by having access to the template text. Such person can't create a filter or any object with new methods, he can only use what has been provided by a programmer in the context. If the context provides an object, in which, there is a method doing something dangerous (like method e.g. getTheClass which returns the class object) it's the programmer's fault, not the problem with the pebble engine, am I right?

In other words, when you use pebble you may provide some objects, strings etc., but all of them extends object in java, so they expose access to getClass() although you didn't provide it explicitly, but if you put some dangerous method in these objects it's your fault.

Br,
Jacek

hectorlf:
If @cjbrooks12 already depended on this, shouldn't we be providing a workaround for his use case? Even if this is a security issue, we're breaking semantic versioning.

cjbrooks12:
I control the object that I'm getting the class from, so I can definitely work around this. But to not break semver, maybe we just include a flag in the PebbleEngine to turn this behavior on/off? It should probably be disabled by default, but make it enabled by default in the next major version

ebussieres:
Done via cd3a821235d35f8016123bd9886615e046b704de. Flag's gonna be false in v3.0

PR #332, opened 2018-04-14
fix for issue308

satyadas2000:
Error message added "endblock tag should be present with block tag starting line number" , if endblock does not exist with block.


ebussieres:
Please provide a test case for your change

PR #338, opened 2018-05-04
Makes logic with allowGetClass and strictVariables well-defined

cjbrooks12:
`allowGetClass` flag was always returning null or throwing ClassAccessException, even when it should have been allowed. This PR fixes that bug, and tests that it conforms to the following truth table (with the prior results also shown):

| AllowGetClass | Strict Variables | Access Type | Expected Result | Prior Result |
| ------------- | ---------------- | ----------- | --------------- | ------------ |
| true          | false            | property    | allowed         | null         |
| true          | false            | method      | allowed         | null         |
| true          | true             | property    | allowed         | throw        |
| true          | true             | method      | allowed         | throw        |
| false         | false            | property    | null            | throw        |
| false         | false            | method      | null            | throw        |
| false         | true             | property    | throw           | throw        |
| false         | true             | method      | throw           | throw        |

### DIFF
--- a/src/main/java/com/mitchellbosecke/pebble/node/expression/GetAttributeExpression.java
+++ b/src/main/java/com/mitchellbosecke/pebble/node/expression/GetAttributeExpression.java
@@ -173,7 +173,7 @@ public class GetAttributeExpression implements Expression<Object> {
                 }
             }
 
-            member = this.reflect(object, attributeName, argumentTypes, context.isAllowGetClass());
+            member = this.reflect(object, attributeName, argumentTypes, context.isAllowGetClass(), context.isStrictVariables());
             if (member != null) {
                 this.memberCache.put(new MemberCacheKey(object.getClass(), attributeName), member);
             }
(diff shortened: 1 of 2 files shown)


ebussieres:
Well I don't think that we should check strict variables to allow call to getclass. We should allow it or not based only on the allowGetClass flag.

cjbrooks12:
I've updated the PR, it now conforms to the following truth table:

| AllowGetClass | Strict Variables | Access Type | Result  |
| ------------- | ---------------- | ----------- | ------- |
| true          | false            | property    | allowed |
| true          | false            | method      | allowed |
| true          | true             | property    | allowed |
| true          | true             | method      | allowed |
| false         | false            | property    | throw   |
| false         | false            | method      | throw   |
| false         | true             | property    | throw   |
| false         | true             | method      | throw   |

ebussieres:
Thanks for you PR ! I'm gonna make a new release as soon as possible to fix this issue. As I see it, it's a blocker one.

PR #353, opened 2018-06-12
First tentative at reducing number of warnings

slandelle:
@ebussieres It would be great the clean up the code and reduce the number of warnings.

Beware that this PR doesn't include #352.


slandelle:
It would be great the address the remaining warnings:
* formatting is not consistant across project. I would recommend to stick a standard, typically IntelliJ defaults, to ease contributions.
* Javadoc errors
* license file header (which is NOT javadoc and shouldn't start with `/**` but `/*`).
* decide what is an API and what is not, and make what is not package private

ebussieres:
@slandelle It's definitely on my TODO list to reformat the code using probably google style formatter.  I'll open an issue with your bullet points

PR #358, opened 2018-06-12
Introduce SpecializedWriter, close #351

### LINKED ISSUE #351: Optimize for in-memory Writers
Pebble's output abstraction is `java.util.Writer` that can only write chars or Strings.
When inserting numerical values, Pebble spends quite some time computing String representations of those values, including allocating and GC'ing those Strings.
Now, when `Writer` actually works in memory and is backed by an `AbstractStringBuilder`, it would be possible to directly append numbers into the underlying buffer and save those Strings allocations.
Such optimization is pretty straightforward for `StringWriter` (it exposes the buffer with `getBuffer`) but it might be nice to introduce an interface for custom `Writer`s.

slandelle:
Motivation:

* `Writer` is the central output abstraction in Pebble. Sadly, it's intended for Strings and chars so numbers have to be turned into their String representation, which cause excessive allocation for number heavy use cases.
* `StringWriter` doesn't provide any optimization for writing numbers, even though the underlying `StringBuffer` does.
* Java's standard `Writer` implementations are synchronized, which can hurt performance.

Modifications:

* Introduce `SpecializedWriter` that provides a way to optimize for writing numbers
* Provide `StringWriterSpecializedAdapter` that's used internally when user provided a `StringWriter`
* Introduce `PooledSpecializedStringWriter` for reusable in memory `StringWriter`s

Result:

Better performance

### DIFF
--- a/src/main/java/com/mitchellbosecke/pebble/extension/writer/PooledSpecializedStringWriter.java
+++ b/src/main/java/com/mitchellbosecke/pebble/extension/writer/PooledSpecializedStringWriter.java
@@ -1,0 +1,86 @@
+package com.mitchellbosecke.pebble.extension.writer;
+
+import java.io.Writer;
+
+/**
+ * A ${@link SpecializedWriter} that's pooled in a ${@link ThreadLocal}.
+ * It's backed by a ${@link StringBuilder} so it's not threadsafe but doesn't involve synchronization.
+ * Beware that it has some limitations:
+ * <ul>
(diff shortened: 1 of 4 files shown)


slandelle:
Current 3.0 head:

```
Pebble.benchmark  thrpt   50  32341.977 ± 258.331  ops/s
```

With `PooledSpecializedStringWriter`:

```
Pebble.benchmark  thrpt   50  37648.746 ± 230.453  ops/s
```

hence ~16% boost

ebussieres:
@slandelle Nice job thanks !

PR #368, opened 2018-07-09
Fixes issues resolving overloaded methods. Closes #367.

### LINKED ISSUE #367: Strange/chaotic/incorrect method resolution with overloaded method signatures
I have been noticing some strange behavior when Pebble attempts to resolve methods that are overloaded. I've added a test case [on this branch](https://github.com/cjbrooks12/pebble/blob/method-resolution-bugs/pebble/src/test/java/com/mitchellbosecke/pebble/bugs/OverloadedMethodErrorTest.java) demonstrating some of the strange behavior. 

Here's my general findings. All are based on a single method name `testMethod` in a model class, and all methods have accept a single argument. There are multiple different versions of this method where the type of the argument changes, but the method name does not.

- When both the primitive and boxed versions exist, it seems like the Boxed version is preferred. I would expect the non-boxed version to be preferred.
- When a more generic type exists, it is preferred to the specific type. This goes all the way to the base case where a method with an argument of type `Object` is preferred to all other types and all method calls go to that one method.
- When two methods both extend the same base class, and a method for the base class doesn't exist, I frequently get either a `ClassCastException` or an `IllegalArgumentException` (argument type mismatch).
	- For this situation, by removing any one method, any of them can successfully be called. Having both methods but only calling one of them also seems fine, so I think this particular issue is related to caching. For reference, I've added a few log statements in `DefaultAttributeResolver` that show the selected method accepts a different parameter than what is passed to it
- All of the above behavior seems very chaotic. Adding or removing an overloaded method from the model may make the primitive method selected over the boxed version, and may also cause a `ClassCastException` instead of an `IllegalArgumentException` as it is currently set up to do. This makes me think that the strategy for selecting a method is not deterministic when it probably should be.

cjbrooks12:
# Changes

Searches matching methods for best possible match rather than returning the first one that works. Adds member parameter types to MemberCacheKey so overloaded methods get resolved from cache correctly, fixing issues with ClassCastException.

# Performance:

Run with [PebbleTemplates/pebble-performance-test](https://github.com/PebbleTemplates/pebble-performance-test) on my MacBook Pro, 4-core 2.8 GHz CPU, 16GB memory.

**Current 3.0**
34335.014766 (±1423.690763) ops/s

**Including these changes**
34743.663107 (±533.180988) ops/s

**Total Change**
+1.19% (-62.55%) ops/s

### DIFF
--- a/pebble/src/main/java/com/mitchellbosecke/pebble/attributes/DefaultAttributeResolver.java
+++ b/pebble/src/main/java/com/mitchellbosecke/pebble/attributes/DefaultAttributeResolver.java
@@ -26,24 +26,7 @@ public class DefaultAttributeResolver implements AttributeResolver {
     if (instance != null) {
       String attributeName = String.valueOf(attributeNameValue);
 
-      Class<?>[] argumentTypes;
-      if(argumentValues != null) {
-        argumentTypes = new Class<?>[argumentValues.length];
-
-        for (int i = 0; i < argumentValues.length; i++) {
-          Object o = argumentValues[i];
-          if (o == null) {
-            argumentTypes[i] = null;
-          }
-          else {
-            argumentTypes[i] = o.getClass();
-          }
-        }
-      }
-      else {
-        argumentTypes = new Class<?>[0];
-      }
-
+      Class<?>[] argumentTypes = this.getArgumentTypes(argumentValues);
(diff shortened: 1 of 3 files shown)


ebussieres:
Good job. I'll take a look later this week. Thanks for your work

PR #387, opened 2018-08-17
Make IF operator more tollerant to argument type

badbob:


ebussieres:
- [ ] Update documentation

ebussieres:
Soon ... I'm out on vacation. I'll do it when I return

Issue #247, opened 2017-01-04
twig compatibility matrix documentation


jochen777:
Request for a compatibility matrix with twig, like here:

https://github.com/justjohn/twig.js/wiki/Implementation-Notes

That would be helpful for me and maybe others... Thank you!

jochen777:
https://pebbletemplates.io/twig-compatibility/ -> perfect!! Thank you!

PR #411, opened 2018-10-25
Sets up Pebble documentation site using Orchid

cjbrooks12:
This PR adds a new maven project in `docs/` which will generate a documentation site and deploy it to Github Pages, using [Orchid](https://orchid.netlify.com/). The site includes a wiki (migrated from the current Github Wiki) and Javadocs documentation, and also includes a basic full-text search of site content. You can [preview](https://cjbrooks12.github.io/pebble/) the site here. It's pretty basic overall, using one of the built-in themes, but I'm more than happy to make any changes as needed.

Here's a brief rundown of the new functionality (will still need to be integrated into the current release process):

**Commands** (run from `docs/` directory)

- `mvn orchid:serve` - build the site and serve it locally on http://localhost:8080. Changes to site content will rebuild the site. 
- `mvn -Denvironment=prod orchid:deploy` - build the site and deploy it to Github Pages. Requires an API token with push access to the repo, set as `githubToken` in the environment variables. The `environment` project property will optimize the build for production.

**Directory Structure**

- `docs/src/orchid/resources` - The main root for site content.
  - `site/src/orchid/resources/pages` - Arbitrary static pages
  - `site/src/orchid/resources/wiki` - A `summary.md` file in here declares the pages in the wiki, which are other files relative to this directory. They will automatically be linked to each other in the order they are defined in the summary file.
  - `site/src/orchid/resources/homepage.md` - The content for the front page of the site.
  - `site/src/orchid/resources/config.yml` - Configuration for things like site menus, and theme/plugin options. 

**Deployment**

Orchid builds the site into the `docs/target/docs/orchid` directory. It has a publication mechanism in it to deploy to Github Pages after a successful build, which creates a temporary directory to do all the git stuff and shouldn't break anything when deploying. It's set up to clean out and force push to the branch, but it has options to maintain history, and also to push each version to a new subdirectory to maintain historical site versions if that is desired. 

I've never worked with Maven so I don't know if this is the correct way to set up the docs project, but everything's working great using the Orchid Maven plugin while in the `docs/` directory. Huge thanks to @bjansen for creating the Maven plugin for this!


bjansen:
Incidentally, since Orchid is generating the site in `target`, this means that the site is contained in the final jar. If we set the packaging to `zip`, we could very easily have an offline version of the site :)

bjansen:
The website feels very slow, but that's because there are HTTP redirects:

* **https**://cjbrooks12.github.io/pebble/wiki/operator/logic redirects to **http**://cjbrooks12.github.io/pebble/wiki/operator/logic/ (notice the trailing /)
* **http**://cjbrooks12.github.io/pebble/wiki/operator/logic/ redirects to **https**://cjbrooks12.github.io/pebble/wiki/operator/logic/

Maybe it's an issue with Github pages though.

cjbrooks12:
All the urls are absolute, so it's not quite set up to just host and view it locally. But I have thought about having the Wiki plugin also generate a pdf with the wiki content, similar to Gitbook, though it would take quite a bit of work. But it's easy enough to clone the whole repo and run the Orchid server on localhost for offline viewing.

cjbrooks12:
I think the redirect speed was just a hiccup, maybe with your computer or with Github's servers. I just checked it out and they are imperceptible. I never even realized it had those redirects before, I've never experienced any poor performance.

bjansen:
image

I just tested again, I'm still seeing the problem :/

cjbrooks12:
screen shot 2018-10-25 at 10 30 56

I have no idea what's going on there for you... is the same thing happening for other GhPages sites?

bjansen:
Weird, when I load the same page a second time, it's super fast. I'll blame the proxy we have at work :)

docs/pom.xml
```diff
@@ -0,0 +1,98 @@
...
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>io.pebbletemplates</groupId>
+        <artifactId>pebble-project</artifactId>
+        <version>3.0.6-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>docs</artifactId>
```
ebussieres: Please add the following or it won't deploy to sonatype

```
<name>Pebble docs</name>
<description>Pebble documentation</description>
<url>http://pebbletemplates.io</url>
```

ebussieres: Please set a packaging too or it's gonna generate a jar file which kinda sucks:

```
<packaging>pom</packaging>
```

ebussieres: Please add the following or it won't deploy when releasing to sonatype 

```
<name>Pebble docs</name>
<description>Pebble documentation</description>
<url>http://pebbletemplates.io</url>
```

cjbrooks12:
> @cjbrooks12 How much time does it take for indexing ? Try to run mvn orchid:serve and it's kinda stuck on site indexing. Maybe I'm missing something ?

The first time you run it, it might take a bit longer as it downloads and caches a few jars to hook Orchid into the Javadoc CLI tool (a clean run just took me about 10 seconds, subsequent runs take about 5). If it's taking much longer than that, something else must be breaking. What OS are you running it on?

ebussieres:
Well after 5 minutes, it's still indexing. I'm on Windows 10. I can try on Ubuntu too

cjbrooks12:
Yeah, thats definitely not right. I develop on a Mac and only really have AppVeyor tests for running on Windows (which seem to be passing fine), but I'll try pulling this open on a Windows VM and see if I can figure out what's going on.

ebussieres:
Tried on Ubuntu and everything worked fine. So yeah there's probably a bug with Windows

ebussieres:
Just published it here : http://pebbletemplates.github.io/pebble/

ebussieres:
It would be great to have the posibility to add a fork me to github ribbon in generated pages. Do you think that's something which is easy to achieve ?

cjbrooks12:
Adding a Github Ribbon is totally possible. It is currently set up using the default Editorial theme with no customization, but any of theme's resources [from here](https://github.com/JavaEden/Orchid/tree/dev/themes/OrchidEditorial/src/main/resources) can be overridden by creating the same file locally. You're probably going to want to override `templates/layouts/index.peb` and add anything you want in there. 

If you want to add additional CSS or JS, you can add it [as described here](https://orchid.netlify.com/wiki/user-manual/content/assets#extra-css-and-js) to the `theme` object in `config.yml`. The referenced assets will be compiled and automatically added to all pages.

In addition, I just added the following features:

- List of contributors. I tried to include everyone who has contributed code since 3.0. Contributors can be added to the list in `data/contributors.yml`.
- Changelogs, using the [Orchid Changelog plugin](https://orchid.netlify.com/plugins/OrchidChangelog). I pulled release notes from the current releases since 3.0. New versions can be added as new files in `changelog/`. it's the `name` property in the file that matters, the filename is irrelevant.
- Automated Github Releases updates. This will require a changelog entry with a version matching the Maven project version, and deploys at the same time as publishing to Github Pages using `mvn orchid:deploy`.
- Twig compatibility matrix, to close out #247. The data for the matrix comes from `data/twig-compatibility`.

ebussieres:
@cjbrooks12 Should we use OrchidAll instead of specifying each dependancy ?

```
           <dependencies>
              <dependency>
                <groupId>io.github.javaeden.orchid</groupId>
                <artifactId>OrchidAll</artifactId>
                <version>${orchid.version}</version>
              </dependency>
            </dependencies>
```

bjansen:
I believe `OrchidAll` pulls way too many jars for our needs.

PR #412, opened 2018-11-04
Improvements in the website

bjansen:
* Described what Pebble is in the home page
* Added a favicon
* Documented long values + string interpolations


bjansen:
On a somewhat related note:

* does Pebble have an official logo? Is this [this one](https://avatars3.githubusercontent.com/u/17010790?s=200&v=4)? 
* what's the official file extension for templates? I've seen `.peb` a lot in `src/test/resources`, but the Spring Boot module is looking for `.pebble` files...

ebussieres:
Yeah you got it right for the official logo 

You're right for the pebble extension. We should stick only to one. I personally prefer .peb. What do you think ?

bjansen:
I'm also used to `.peb`, but if we change the default extension in the Spring Boot module, this will result in a breaking change.

bjansen:
Hang on, I'll add another commit to this PR to change the favicon in the generated site :)

bjansen:
Here's how the favicon looks on my retina screen:

bjansen:
I think I'm done :)

Issue #391, opened 2018-08-21
Any plans to support Java 9 modules (Jigsaw)?


mtdowling:
Are there any plans for this project to support Java 9 modules?

ebussieres:
We've just moved to java 8 as minimum requirement so no, not in the short term. 

What's your need with java 9 modules ?

mtdowling:
Adding support for modules allows people to use Pebble with tools like [jlink](https://docs.oracle.com/javase/9/tools/jlink.htm) that can create a standalone executable that contains only the classes and parts of the JRE that are used in their application.

To support it in a way that doesn't require Java 9 but still supports modules is an interesting question. I know that Jackson, Guava, and other popular projects are trying to to figure this out too.

colesico:
add at least an java9 Automatic-Module-Name  to manifest

ebussieres:
@colesico Never tried java9 module. You can submit a PR to add that to the manifest if you need it

zimmi:
I opened https://github.com/PebbleTemplates/pebble/pull/416 to add automatic module names. Please feel free to comment, since the module name is supposed to be permanent.

zimmi:
@ebussieres Sorry for nagging, but can you take a look at PR https://github.com/PebbleTemplates/pebble/pull/416? An official, stable module name would greatly help those projects using pebble on the module path, right now they have to make up a name or use the generated one, which will likely break in the future.

Issue #311, opened 2017-12-13
Is it dying?


decebals:
Hi,

I see no activities (interest) for this project in the last period of time? They are many PR and issues with no resolution or comment. 
@mbosecke Do you intent to continue working on this project? Any information from maintainers about the next steps/releases is very welcome.

DirkLachowski:
Ping.

deelo55:
Seems pretty dead...

hectorlf:
Sorry, for the delay, guys. Mitchell has some IRL things to attend to, so Eric and I will try to do some of the housekeeping. Please be patient, as we are just the reserves.

hectorlf:
Not me, that's for sure.

DirkLachowski:
@hectorlf Hey hector, glad you have stepped in. Just to understand you right: Is this a confirmation that the project is not dying but that it's only a resource problem (i would be fine with that) or will a new maintainer be needed sometime?

deelo55:
Given there has been no response on this for over 3 months I assume that it is actually dying. Should we fork the project? Is anyone interested in maintaining this?

hc-codersatlas:
Would be happy to fork / maintain this. We use it for quite a few private projects. Anyone else interested?

cjbrooks12:
I would be interested in helping maintain a fork with you @hc-codersatlas, @deelo55. I've been using Pebble as the main template engine in my new static site generator, [Orchid](https://github.com/JavaEden/Orchid), and the future of Pebble is pretty important for its success. 

I've gone ahead a set up a fork under Orchid's organization and would love to host the new fork there indefinitely. I will invite anyone who wishes to be an admin in that project, and will get it set up to release to Bintray as well (under a different groupId, obviously).

https://github.com/JavaEden/pebble

cjbrooks12:
Alright, I've now got Pebble moved over to my organization @JavaEden and set up on Bintray, with Maven Central coming soon (see checklist below for status). I will go through all the open PRs and request they be moved to the new repo. Please take new issues there as well. In addition, I would like to pick back up discussions around releasing Pebble 3.0, and welcome you all to join the conversation on Gitter. 

**Links**
- https://github.com/JavaEden/pebble
- https://bintray.com/javaeden/Eden/Pebble
- https://gitter.im/JavaEden/pebble

**Maven Central Checklist**
- [ ] Project needs to be synced to jCenter
[...]

PR #416, opened 2018-12-05
Add automatic module names

zimmi:
This is a first step towards https://github.com/PebbleTemplates/pebble/issues/391.

The primary goal for module names should be uniqueness, since they all share a namespace.
The [official naming strategy recommendation](http://mail.openjdk.java.net/pipermail/jpms-spec-experts/2017-May/000687.html) says:

> Strongly recommend that all modules be named according to the reverse
    Internet domain-name convention.  A module's name should correspond
    to the name of its principal exported API package, which should also
    follow that convention.  If a module does not have such a package, or
    if for legacy reasons it must have a name that does not correspond to
    one of its exported packages, then its name should at least start
    with the reversed form of an Internet domain with which the author is
    associated.

What makes this project special:
1. The _principal exported API package_ is  `com.mitchellbosecke.pebble` but the registered domain is `pebbletemplates.io`. Reading https://github.com/PebbleTemplates/pebble/issues/311, pebble might have such a "legacy reason" the recommendation speaks of. I therefore propose `io.pebbletemplates.pebble` as the module name. If the package name gets changed in the future it will be more intuitive, but I feel it's wrong to use @mbosecke s name without his input. I considered only `io.pebbletemplates`, but that felt too generic.
2. The spring and spring boot modules support different spring versions. Since spring doesn't change its package with new versions and it's therefore impossible to have multiple spring versions on the module path at the same time, I gave all modules the same module name `io.pebbletemplates.pebble.spring`. Also makes upgrading easier.
3. The spring boot modules have the very long package name `com.mitchellbosecke.pebble.boot.autoconfigure`. I therefore shortened the module names to `io.pebbletemplates.pebble.boot`.

Does anybody have any strong feelings about this?

Edit: If there are no objections, this is ready to merge.

### DIFF
--- a/pebble-spring/pebble-spring-boot-2-starter/pom.xml
+++ b/pebble-spring/pebble-spring-boot-2-starter/pom.xml
@@ -38,4 +38,19 @@
       <scope>test</scope>
     </dependency>
   </dependencies>
+  
+  <build>
+    <plugins>
+      <plugin>
+        <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
+          <archive>
+            <manifestEntries>
+              <Automatic-Module-Name>io.pebbletemplates.pebble.boot</Automatic-Module-Name>
+            </manifestEntries>
+          </archive>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
 </project>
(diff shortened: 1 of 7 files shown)


ebussieres:
Personnaly, I'd stick with the following module names

1. io.pebbletemplates for pebble project
2. io.pebbletemplates.spring for pebble-spring project
3. io.pebbletemplates.springboot for pebble-spring-boot-starter project. 

What do you think ?

zimmi:
Thanks for taking a look. That's fine as well, the important bit is the reverse domain prefix to ensure uniqueness. I'll go ahead and change them

electrotype:
Thanks for that PR!

Issue #378, opened 2018-07-30
Component based html page   (question)


colesico:
I plan to create pages based on reusable, nested components.
For example, a simple button component I can define with the macro (as mentioned in the documentation):

```
{% macro button (label)%}
<input type = "button" value = "{{label}}" />
{% endmacro%}

```
And somewhere on the page I can specify two buttons:

`{{button ("Ok")}}  <br/> {{button ("Cancel")}}`

It is clear how to do this.

But what about  complex component, where the parameters values must be not just text strings  but whole blocks of code, for example:

I made the "Dialog" component that has a "header" and a "body" parts:

```
{% macro dialog (header, body)%}
	<div>
		<div id = "header"> {{header}} </ div>
		<div id = "body"> {{body}} </ div>
	</ div>
{% endmacro%}
```

And somewhere on the page I use it twice:  

for the first use:

` {{dialog ("My header", body ="My body")}}   `

and for the second use, the value of the "body" parameter must be not just string but the code block:

```
  <div>
     {{button ("Ok")}}  <br/> {{button ("Cancel")}}
  </ div>
```
  
Also page can contains not just one component like the "Dialog" but multiple different components with  blocks of code as the parameters values

Thnx!

cjbrooks12:
To use a macro, it must be defined in the same template, or else [imported](https://github.com/PebbleTemplates/pebble/wiki/import) from another template.

Currently it is not possible to pass an HTML block to a macro, but there has been talk of adding an `embed` tag (#319) that would allow similar functionality.

cjbrooks12:
Include statements are evaluated dynamically, so one solution would be to pass a string to your macro, where that string defines a template name to include. Not ideal, but it would certainly allow for this kind of reuse.

```jijna
{% macro dialog(header, body, vars) %}
<div>
<div id="header">{% include header with {"vars": vars} %}</div>
<div id="body">{% include body with {"vars": vars} %}</div>
</div>
{% endmacro%}

{{ dialog('includes/dialog/header.peb', 'includes/dialog/body.peb', {"title": "My Title"}) }}
```

Issue #319, opened 2018-02-03
Embed tag


deelo55:
At the moment, there is no implementation of the Twig embed tag. This really hinders the ability to reuse code as include tags cannot contain any blocks.

Additionally the set tag in twig allows arbitary blocks of html:

`{% set foo %}     <div id="pagination">         ...     </div> {% endset %}`
--
This would also be a useful feature to implement.

Is there anyone actively working on this project? Its a great templating engine....would be a shame to see it die.

hectorlf:
Could you provide a use case for the two issues? I've had a look at the embed tag in Twig and I have a hard time figuring out what would I use it for. So, please, an example of a layout you're trying to achieve would be great. And the same goes for the set tag.

cjbrooks12:
I've actually used the `embed` tag heavily in PHP Twig, and would love to see it implemented here. The way I have used it is in a CMS setting, to implement a 'wrapper" or "container" element that can be reused horizontally across many different layouts. For example:

**layouts/page.peb**
```twig
{% extends 'layouts/base.peb' %}
{% block main %}
    <div id="main">
    {% embed 'includes/container.peb' %}
        // uses default header and content, but overrides footer for static page CTA
        {% block pageFooter %}{% endblock %}
    {% endembed %}
    </div>
{% endblock %}
{% endextends %}
```

**layouts/post.peb**
```twig
{% extends 'layouts/base.peb' %}
{% block main %}
    <div id="main">
    {% embed 'includes/container.peb' %}
        // uses default footer and content, but overrides header to show breadcrumbs, tags, author, etc.
        {% block pageHeader %}{% endblock %}
    {% endembed %}
    </div>
{% endblock %}
{% endextends %}
```

**includes/container.peb**
```twig
{% block pageHeader %}{% endblock %}
{% block pageContent %}{% endblock %}
{% block pageFooter %}{% endblock %}
```

Compare this to the equivalent without the 'embed' tag:

**layouts/page.peb**
```twig
{% extends 'layouts/base.peb' %}
{% block main %}
    <div id="main">
        <div id="container">
            {% include 'includes/container/header.peb' %}
            {% include 'includes/container/content.peb' %}
            // you can override footer here, but there's no way to know what the "normal" thing to do here is, plus you have to manually bring over the wrapper markup
        </div>
    </div>
{% endblock %}
{% endextends %}
```

**layouts/post.peb**
```twig
{% extends 'layouts/base.peb' %}
{% block main %}
[...]

Issue #318, opened 2018-02-02
Blocks in include files


deelo55:
Is it possible to have a block in an include file.
For example:

head.peb:
```
<head>
<title>{% block title %}{% endblock %}</title>
</head>
```

parent.peb:
```
<!DOCTYPE html> <html> {% include "include/head" %} 
<body> 
{% block content %}{% endblock %}  
</body>
 </html>
```

child.peb:
```
{% extends "parent.peb" %}
{% block title %}My Title{% endblock %}

{% block content %}HELLO{% endblock %}
```

colesico:
As far as I know this is impossible due the bad design. The included file has its own independent blocks hierarchy.
You can pass a string variable as the title value

Issue #224, opened 2016-09-04
Passing the content of a block section to an included fragment


electrotype:
Let's say I have a main `layout.html` template which has a "body" block and also includes a `header.html` fragment. Every concrete pages extend `layout.html` and define their custom "body" `block.

My problem is that the `header.html` template displays a menu, but some concrete pages need to add a submenu, just below it!

Is it possible for the main `layout.html` template to "retrieve" such submenu content from a concrete page and _pass it_ to the included `header.html`?

A dummy example would be :

 `layout.html` :

```
...
{% set subMenu = {% block subMenu %}{% endblock %} %}
{% include "./header.html" with {"subMenu", subMenu} %}
{% block body %}{% endblock %}
```

 `concretePage.html` :

```
{% extends "./layout.html" %}
...
{% block subMenu %}
    <div>concrete page submenu</div>
{% endblock %}

{% block body %}
    <div>concrete page body</div>
{% endblock %}
```

 `header.html` :

```
<div>main menu</div>
{% if subMenu is not empty %}
    {{subMenu | raw}}
{% endif %}
```

Is something similar to this possible?
Thanks!

tboyce021:
The way I've worked around this is to wrap it in a macro:
```
{% macro subMenu() %}
    {% block subMenu %}{% endblock %}
{% endmacro %}

{% set subMenu = subMenu() %}
```
You then run into the issue of the macro not having the same scope as the template and have to work around that though.

PR #434, opened 2019-03-25
Embed tag for overriding blocks in included templates

cjbrooks12:
This PR adds an `embed` tag, which works like [Twig's `embed` tag](https://twig.symfony.com/doc/2.x/tags/embed.html). It effectively includes a template but allows that template's blocks to be overridden in the host template within the body of the `embed` tag.

This will address a number of open issues, it's clearly a highly-desired feature: 
- #224
- #318
- #319
- #378

My initial work on this seems very promising. I am opening this PR now so the community knows it is coming soon, and to get more eyes on it and make sure it's being implemented correctly.

**To-Do:**

- [x] More tests, especially around error-handling, and verifying it doesn't cause strange behavior when used in a template that uses other block-inheritance mechanisms (`extends`, `block`, `parent()`, `block(name)`)
- [x] Documentation
- [x] Verify Pebble's `embed` behavior matches that of Twig


cjbrooks12:
I've added a bunch more tests, and am comparing them to how Twig would render them, and I can't find any differences yet. Would love to get some more eyes on the PR, especially from people who have used Twig's `embed` tag in real-world usage. 

But besides that, I just need to add some documentation to the wiki and I think this should be ready-to-go.

cjbrooks12:
Alright, I think this PR is ready to be merged at this point

Issue #441, opened 2019-05-20
JDK9 illegal reflective action operation warning


NatHerencia:
Starting with JRE9, when executing "((AccessibleObject) result).setAccessible(true);" in MemberCacheUtils.java:83, the JRE throws a warning telling that, in future versions, that access won't be allowed.

This is related with "new" JDK modules. If a module doesn't explicitly allow another module to access it by Reflection, then the JDK currently throws just a warning **but in the future it won't allow access**.

In order to see the issue, execute MapSyntaxTest.testForTag() method using JDK9 or greater.

`WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.mitchellbosecke.pebble.attributes.MemberCacheUtils (file:/proyectos/repositorios/pebble/pebble/target/classes/) to method java.util.HashMap$Node.getKey()
WARNING: Please consider reporting this to the maintainers of com.mitchellbosecke.pebble.attributes.MemberCacheUtils
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release`

Specifically, the warning is related to an old JDK bug ( https://stackoverflow.com/questions/12037489/cant-access-to-java-util-hashmapentry-with-modifiers-public-final ) and it can be solved by making a new implementation of Map.Entry interface because Map module, obviously, doesn't allow specifically to Pebble to access it by Reflection.

ebussieres:
Thanks ! Gonna be in next release

zimmi:
Hi! I noticed that this is not only a problem for `HashMap`s but also `Arrays::asList`, so I opened a PR which handles this problem (hopefully) for the more general case of inaccessible collection implementations: https://github.com/PebbleTemplates/pebble/pull/443.

PR #443, opened 2019-05-27
More comprehensive fix for illegal reflective access under jdk 9+

zimmi:
While looking into #441 I noticed that the same problem occurs for the private `List` implementation returned by `Arrays::asList` (the `size()` method can't be called, see `LogicTest::testListSize`). That's a general problem for unexported / unopened collection types. The fix is to search these well known public classes for members first. This approach also avoids copying potentially large collections.

~Adding `--illegal-access=deny` to unit tests will fail the build for these errors to avoid regressions in the future.~ Doesn't work, see https://github.com/PebbleTemplates/pebble/pull/443#issuecomment-496643233.


ebussieres:
@zimmi Did you run the performance benchmark to see if there's a difference ? https://github.com/PebbleTemplates/pebble-performance-test

zimmi:
@ebussieres Nope, will do that now.

zimmi:
Running on `openjdk 12.0.1 2019-04-16`

#### master
    Pebble.benchmark  thrpt   50  24635.207 ± 180.448  ops/s

#### this PR
    Pebble.benchmark  thrpt   50  24335.089 ± 253.823  ops/s

ebussieres:
Can you update .travis.yml file to test with multiple jdk ?

zimmi:
The failures are because https://github.com/PebbleTemplates/pebble/pull/445 is not yet merged. Do you want me to add those changes in here and delete the other PR?

ebussieres:
Yes please merge the two PR together. 

Thanks

zimmi:
It is done.

ebussieres:
Thanks

zimmi:
Nice, thanks for merging! And thanks to @NatHerencia for the hard work of finding out the root issue!

PR #450, opened 2019-07-06
Remove unnecessary variable 'state' in LexerImpl by using Stack

yaboong:
In my opinion, the 'state' variable in LexerImpl seems unnecessary. You can remove them by using Stack data structure to check the current state. (Replaced LinkedList to Stack to make it clear that the 'states' is managed by Stack)

All the test cases have passed.

### DIFF
--- a/pebble/src/main/java/com/mitchellbosecke/pebble/lexer/LexerImpl.java
+++ b/pebble/src/main/java/com/mitchellbosecke/pebble/lexer/LexerImpl.java
@@ -22,6 +22,7 @@ import java.util.Collection;
 import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
+import java.util.Stack;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
@@ -70,9 +71,7 @@ public final class LexerImpl implements Lexer {
    * The state of the lexer is important so that we know what to expect next and to help discover
    * errors in the template (ex. unclosed comments).
    */
-  private State state;
-
-  private LinkedList<State> states;
+  private Stack<State> states;
 
   private enum State {
     DATA, EXECUTE, PRINT, COMMENT, STRING, STRING_INTERPOLATION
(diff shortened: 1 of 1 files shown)


zimmi:
`java.util.Stack` is effectively deprecated in favor of `Deque`, here is what the [javadoc](https://docs.oracle.com/en/java/javase/12/docs/api/java.base/java/util/Stack.html) has to say:

> A more complete and consistent set of LIFO stack operations is provided by the Deque interface and its implementations, which should be used in preference to this class. For example:
   `Deque<Integer> stack = new ArrayDeque<Integer>();`

yaboong:
@zimmi 
Now I understand that Pebble is using LinkedList instead of Stack. Thanks for letting me know.
However, the 'state' variable still seems unnecessary for me because you can use the peek () method to check the current state. Is it okay to create another PR to remove the 'state' variable from the LexerImpl?

zimmi:
I'm not a maintainer of this repo, just a happy user of pebble. :)
That said, your change looks fine to me. Just replace the `Stack` with `ArrayDeque`, the previous `LinkedList` is probably also less efficient. Most methods of `Stack` are needlessly synchronized, that's one of the reasons it shouldn't be used anymore. The [javadoc](https://docs.oracle.com/en/java/javase/12/docs/api/java.base/java/util/ArrayDeque.html) of `ArrayDeque`says:

>This class is likely to be faster than Stack when used as a stack, and faster than LinkedList when used as a queue. 

I'd just update this PR, but you might want to wait for one of the maintainers to make the call.

yaboong:
@zimmi 
All right then I'll wait until the maintainer confirms this PR. Thanks for your feedback. It helped me to understand Java better.

yaboong:
@zimmi 
Okay then I'll close this and change the Stack to ArrayDequeue to create another PR. Thank you for your feedback. It helped me to better understand Java :)

yaboong:
@zimmi 
Thank you for your feedback. As you said, I will wait for the maintainer to confirm this PR :)

zimmi:
Instances of `LexerImp` are never called from multiple threads, [each thread creates its own instance](https://github.com/PebbleTemplates/pebble/blob/5c764071fc4df8475b985d87f58dfe8e8a342fde/pebble/src/main/java/com/mitchellbosecke/pebble/PebbleEngine.java#L154). So the synchronization is useless. That aside, just using thread safe data structures is not enough to achieve thread safety of a whole program. See [this SO question](https://stackoverflow.com/questions/1386275) for the concrete case of `Vector` / `Stack`.

ebussieres:
Should you use Deque instead of Stack ?

yaboong:
This could be debatable. Deque could be better if you're looking at performance, but Stack is better in terms of semantics.

Deque can be used like Stack at times, and sometimes like Queue but I think this is a double edged sword. Stack, on the other hand, is just Stack. A Stack cannot be used just like a Queue even if it extends Vector.

I think it's a matter of choice.

slandelle:
IMHO, no, choice is not debatable: Stack is obsolete and Dequeue’s performance is better.

yaboong:
Alright then is it okay to close this PR and make another PR to change all the Stack to Deque?

slandelle:
Why not edit this one? You just to update your fork.

ebussieres:
No need to close this PR. Just commit and push to your branch and the PR will bu updated

yaboong:
It's done. Thanks.

Issue #470, opened 2019-09-02
Whitespace Control with for loop and nested if statement


nward1234:
I am not able to control the vertical spacing/whitespace when I have a template that has an `if` statement inside a `for` loop that ends up skipping some of the items in the list. I have the Pebble Engine configured to **disable New Line Trimming** since I see no way to get the vertical spacing that I want with New Line Trimming enabled. 

I have made a JUnit test to demonstrate the issue. The example template as well as expected template output and actual template output is shown below. I will submit a pull requests with the test code. One test is failing due to this issue. That test is marked with `@Ignore` so that adding the tests does not break the build. 

**Template:**
```twig
text before for loop followed by blank line
{% for item in items %}
{% if item.itemType equals "ITEM_TYPE1" -%}
Item 1
{% elseif item.itemType equals "ITEM_TYPE2" -%}
Item 2
{% endif -%}
{% endfor %}
text after for loop preceded by blank line
```

**Expected Template Output:**
```
text before for loop followed by blank line

Item 1

Item 2

text after for loop preceded by blank line
```

**Actual Template Output:**
```
text before for loop followed by blank line

Item 1

Item 2



text after for loop preceded by blank line
```

This is a big problem for me since I am using Pebble for code generation. This results in generated code such as the following, which is a real problem for generated code. I will be glad to work on fixing this issue in the Pebble code if you confirm that this is an issue. If there is another way to get the desired output in a Pebble template, please let me know.

```
public class SchoolModel implements Serializable {

    private static final long serialVersionUID = 1L;

    private Long id;

    public String identifier;
    
    @NotNull
    private String abbreviation;
    
    
    
    
    
    private LocalTime goHomeNotesDailyCutoffTime;
    
    private LocalDate goHomeNotesStartDate;
    
    private LocalDate goHomeNotesStopDate;
    
    @NotNull
    private String name;
    
    
    
    
    
    
    
    public String getIdentifier() {
        return this.identifier;
    }

    public void setIdentifier(String identifier) {
        this.identifier = identifier;
    }

    public Long getId() {
        return id;
    }

    public void setId(Long id) {
        this.id = id;
    }
    
    public String getAbbreviation() {
        return this.abbreviation;
    }

    public void setAbbreviation(String abbreviation) {
        this.abbreviation = abbreviation;
    }
    
    
    
    
    
    public LocalTime getGoHomeNotesDailyCutoffTime() {
        return this.goHomeNotesDailyCutoffTime;
    }
```

[...]

PR #471, opened 2019-09-02
Whitespace Control Template Tests with New Line Trimming Disabled

nward1234:
I added tests for whitespace control for templates when **New Line Trimming is disabled**. Also added **assertj** and **commons-io** Maven dependencies for use in the tests. Commons IO is used to compare template output to expected output that is loaded from a file. Loading the expected template output from a file is useful for testing multiline template output, which is necessary to test various scenarios. These are a form of Integration Test rather than Unit Tests and so it is practical and reasonable to verify the template output in this way.

Related issue: https://github.com/PebbleTemplates/pebble/issues/470#issue-488221680


nward1234:
I will cancel/close this pull request and submit another request since I have made additional improvements to the tests.

PR #475, opened 2019-09-11
Fix issue of embed tag changing cached templates

cjbrooks12:
Fix issue of embed tag changing cached templates and breaking includes using the same-named template from the cache.

For example, let's say `templateA.peb` embeds `templateX.peb`, while `templateB.peb` just includes it. If template caching is enabled, both the embed and include tag will use the same cached `PebbleTemplateImpl` object. But for the embed tag to work, it needed to change some of the data in that template, so that later on, when it's just included later, the modified template is being included.

By making a shallow copy of the cached template, the embed tag can make the changes it needs to without affecting what's already in the cache.


ebussieres:
@cjbrooks12 Build is failling

cjbrooks12:
Build's fixed

PR #476, opened 2019-09-19
Added StringResolver

axeld:
This allows to replace the toString() method that is used in RawFilter
and EscapeFilter with one that has access to the context, ie. to access
the current locale and other things.

### DIFF
--- a/pebble/src/main/java/com/mitchellbosecke/pebble/extension/escaper/EscapeFilter.java
+++ b/pebble/src/main/java/com/mitchellbosecke/pebble/extension/escaper/EscapeFilter.java
@@ -61,7 +61,12 @@ public class EscapeFilter implements Filter {
     if (inputObject == null || inputObject instanceof SafeString) {
       return inputObject;
     }
-    String input = StringUtils.toString(inputObject);
+    String input;
+    if (inputObject instanceof StringResolver) {
+      input = ((StringResolver)inputObject).resolve(inputObject, self, context, lineNumber);
+    } else {
+      input = StringUtils.toString(inputObject);
+    }
 
     String strategy = this.defaultStrategy;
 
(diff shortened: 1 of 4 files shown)


ebussieres:
I'm relunctant to merge that change. instanceof affects performance a lot and I think it's gonna drop. 

Did you take a look at AttributeResolver interface and add it to the registry ? Maybe you can achieve what you need with that

https://pebbletemplates.io/wiki/guide/extending-pebble/

axeld:
Well, it would be nice if we could actually compare the performance with how it did before, before throwing it away. But if it turns out to be a performance issue, one could still make it optional.

I'd be happy to implement it in another way, I just don't know how. I need to access the context from toString() in order to resolve the current language, mostly. The object I'm using this with is already an `AttributeResolver`.

axeld:
I just found this reference concerning the `instanceof` speed: https://stackoverflow.com/questions/103564/the-performance-impact-of-using-instanceof-in-java
I would not expect using `instanceof` here would make much of a difference.

ebussieres:
@axeld You can run this benchmark https://github.com/PebbleTemplates/pebble-performance-test

axeld:
Hey, I even implemented a unit test! ;-)
Alright, I've replaced the "raw" and "escape" filters now only in my own template engine, and that works fine, too. 

However, there is one thing I missed in my patch: PrintNode also just uses toString() if auto escaping is turned off -- there is currently no way I know of to replace this using an extension. IOW I cannot completely implement my feature without hacking on the core. How should I proceed here?

ebussieres:
I don't think it's gonna be easy to replace PrintNode. 

But do you really need to turn off auto espaping ? In this PR, you just override EscapingFilter and RawFilter

objectProgrammer:
Can this get closed?

Issue #449, opened 2019-07-01
Using an unmodifiable map as context causes errors while evaluating the template


hex-agon:
Currently i'm using pebble with [javalin](https://javalin.io/) and by default, when not specifying a context map, javalin passes a `emptyMap()` to pebble and this causes an `UnsupportedOperationException` to be thrown:
```
java.lang.UnsupportedOperationException: Operation is not supported for read-only collection
	at kotlin.collections.EmptyMap.put(Maps.kt) ~[kotlin-stdlib-1.3.31.jar:1.3.31-release-198 (1.3.31)]
	at com.mitchellbosecke.pebble.template.Scope.put(Scope.java:68) ~[pebble-3.0.10.jar:?]
	at com.mitchellbosecke.pebble.template.ScopeChain.put(ScopeChain.java:96) ~[pebble-3.0.10.jar:?]
	at com.mitchellbosecke.pebble.template.ScopeChain.set(ScopeChain.java:228) ~[pebble-3.0.10.jar:?]
	at com.mitchellbosecke.pebble.node.SetNode.render(SetNode.java:31) ~[pebble-3.0.10.jar:?]
	at com.mitchellbosecke.pebble.node.BodyNode.render(BodyNode.java:43) ~[pebble-3.0.10.jar:?]
	at com.mitchellbosecke.pebble.node.RootNode.render(RootNode.java:30) ~[pebble-3.0.10.jar:?]
	at com.mitchellbosecke.pebble.template.PebbleTemplateImpl.evaluate(PebbleTemplateImpl.java:149) ~[pebble-3.0.10.jar:?]
	at com.mitchellbosecke.pebble.template.PebbleTemplateImpl.evaluate(PebbleTemplateImpl.java:92) ~[pebble-3.0.10.jar:?]
	at io.javalin.plugin.rendering.template.JavalinPebble.render(JavalinPebble.kt:31) ~[javalin-3.1.0.jar:?]
	at io.javalin.plugin.rendering.JavalinRenderer.renderBasedOnExtension(JavalinRenderer.kt:33) ~[javalin-3.1.0.jar:?]
	at io.javalin.http.Context.render(Context.kt:440) ~[javalin-3.1.0.jar:?]
	at io.javalin.http.Context.render$default(Context.kt:439) ~[javalin-3.1.0.jar:?]
	at io.javalin.http.Context.render(Context.kt) ~[javalin-3.1.0.jar:?]
```

[...]

PR #479, opened 2019-09-27
Extract Scope interface

jaspeen:
Changes: 
* Extract interface Scope and move implementation to ScopeImpl. Use interface where possible.
* Add isWritable() method to the Scope
* Add evaluate(Writer writer, Scope scope) method to PebbleTemplateImpl 

After change in https://github.com/PebbleTemplates/pebble/issues/449 i no long can save context between evaluate calls so if template set/update some variable it will not be visible in subsequent execution of evaluate with same map instance. 
In my case it is necessary feature so i proposing this abstraction which also should target cases where context map is not a map itself but some proxy/dynamic content and creating java.util.Map proxy implementation might be dirty solution. 

Important is a way to share same context between evaluate calls somehow. If there is another better  solution im glad to implement it instead.

### DIFF
--- a/pebble/src/main/java/com/mitchellbosecke/pebble/template/ScopeImpl.java
+++ b/pebble/src/main/java/com/mitchellbosecke/pebble/template/ScopeImpl.java
@@ -1,0 +1,74 @@
+/*
+ * This file is part of Pebble.
+ * <p>
+ * Copyright (c) 2014 by Mitchell Bösecke
+ * <p>
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+package com.mitchellbosecke.pebble.template;
+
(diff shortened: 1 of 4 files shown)


pebble/src/main/java/com/mitchellbosecke/pebble/template/ScopeImpl.java
```diff
@@ -0,0 +1,74 @@
...
+  /**
+   * A "local" scope ensures that the search for a particular variable will end at this scope
+   * whether or not it was found.
+   */
+  private final boolean local;
+
+  /**
+   * The map of variables known at this scope
+   */
+  private final Map<String, Object> backingMap;
```
nward1234: Since this is a map of known variables, why isn't it names something like `knownVariables` or `variablesMap`?

PR #480, opened 2019-10-20
LexerImpl Unit Tests

nward1234:
Added unit tests for the LexerImpl class and did some minor renaming and minor refactoring in LexerImpl to improve clarity (no functional changes -- just adding tests at this time). Also added log4j debug level log statements in LexerImpl, PebbleEngine and LexerImpl that were useful to me when trying to follow the code. Added logback-test.xml to configure the logging level. Disabled the debug logging by default. Also removed Hamcrest Matchers and replaced references to Hamcrest in exception testing code with AssertJ.


ebussieres:
Good job !

Issue #487, opened 2019-11-23
pebble doesn't work on jdk 13


jknack:
On JDK13 a simple initialization, results in an error.

```java
PebbleEngine build = new PebbleEngine.Builder().build();
```

Stacktrace:
```
	java.lang.ExceptionInInitializerError
	at com.mitchellbosecke.pebble.attributes.MemberCacheUtils.<init>(MemberCacheUtils.java:19)
	at com.mitchellbosecke.pebble.attributes.DefaultAttributeResolver.<init>(DefaultAttributeResolver.java:16)
	at com.mitchellbosecke.pebble.extension.core.AttributeResolverExtension.getAttributeResolver(AttributeResolverExtension.java:14)
	at com.mitchellbosecke.pebble.extension.ExtensionRegistry.addExtension(ExtensionRegistry.java:144)
	at com.mitchellbosecke.pebble.extension.ExtensionRegistry.addExtension(ExtensionRegistry.java:77)
	at com.mitchellbosecke.pebble.PebbleEngine$Builder.buildExtensionRegistry(PebbleEngine.java:596)
	at com.mitchellbosecke.pebble.PebbleEngine$Builder.build(PebbleEngine.java:535)
	at io.jooby.pebble.PebbleModuleTest.render(PebbleModuleTest.java:40)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:567)
	at org.junit.platform.commons.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:675)
	at org.junit.jupiter.engine.execution.MethodInvocation.proceed(MethodInvocation.java:60)
	at org.junit.jupiter.engine.execution.InvocationInterceptorChain$ValidatingInvocation.proceed(InvocationInterceptorChain.java:125)
....
Caused by: java.lang.RuntimeException: Could not load unsafe method set
	at com.mitchellbosecke.pebble.attributes.UnsafeMethods.createUnsafeMethodsSet(UnsafeMethods.java:35)
[...]

PR #488, opened 2019-11-25
#487: Add support for jdk13

ebussieres:


slandelle:
Shouldn’t you have one blacklist per JDK version ?

PR #503, opened 2020-04-22
Allow treating all number literals as BigDecimals to avoid NumberFormatExceptions

eXsio:
If a literal is big enough, with current implementation there will be a NumberFormatException thrown during the processing of the Template. There should be an option to treat all number literals as BigDecimals. It will help avoiding this exception and also enforce precise arithmetics when performing calculations.


ebussieres:
I'll need to run some benchmark with these changes. But it sounds good to me

eXsio:
Sure, docs updated. Setting this option to true may indeed cause some small performance degradation (BigDecimals are slower than Longs/Doubles by nature). I'd still keep it for folks that would rather sacrifice a bit of performance but gain 100% precision when doing math.

In my use case I'm using Pebble as an expression evaluation engine (to calculate Excel-like formulas using my own set of Functions and Operators) and this option is vital to me - without it I have to make workarounds by creating custom Functions to explicitly convert String-based args to BigDecimals.

Issue #414, opened 2018-11-12
 Support for Chinese variable syntax?


qiushasha:
### When set a chinese variable with template as this: {{商品名称}}, but cause a exception as bellow,
and I found some issues:
 1. The LexerImpl is not support Chinese variable syntax with  regular pattern as bellow:
  Pattern REGEX_NAME = Pattern.compile("^[a-zA-Z_][a-zA-Z0-9_]*"); 
 2.  The PebbleEngine  could give some APIs for user to implements her/she logic

Exception information:
com.mitchellbosecke.pebble.error.ParserException: Unexpected character [商] ({{商品名称}}:1)
	at com.mitchellbosecke.pebble.lexer.LexerImpl.lexExpression(LexerImpl.java:509)
	at com.mitchellbosecke.pebble.lexer.LexerImpl.lexPrint(LexerImpl.java:366)
	at com.mitchellbosecke.pebble.lexer.LexerImpl.tokenize(LexerImpl.java:181)
	at com.mitchellbosecke.pebble.PebbleEngine.getPebbleTemplate(PebbleEngine.java:157)
	at com.mitchellbosecke.pebble.PebbleEngine.lambda$getTemplate$1(PebbleEngine.java:146)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache.lambda$doComputeIfAbsent$14(BoundedLocalCache.java:2039)
	at java.util.concurrent.ConcurrentHashMap.compute(ConcurrentHashMap.java:1853)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache.doComputeIfAbsent(BoundedLocalCache.java:2037)
	at com.github.benmanes.caffeine.cache.BoundedLocalCache.computeIfAbsent(BoundedLocalCache.java:2020)
..........................................

ebussieres:
I really don't know nothing about Chinese syntax, so it's gonna be hard for me. Would you like to make a pull request for this ?

qiushasha:
en，I will learn some code  and try to implement

nil4u:
@qiushasha 为什么要用中文变量？ 不建议..

yanxiyue:
@qiushasha 为什么要用中文变量？ 不建议...

qiushasha:
@yanxiyue  at some cases, I need to build-in some  variables with chinese alias which can be convenient to user edit his/her template, so I hope PebbleEngine or Lexer give some extension points to implements some logic, but I don't know what to do

[...]

PR #544, opened 2020-10-28
Support all Java identifiers

### LINKED ISSUE #545: All Java identifiers are not supported
Fixed by #544

objectProgrammer:
Fixes #545.

This is needed to call methods or use fields that are not a-z characters, for example éêȩ etc.

These are supported in Java and because Java provides useful patterns for matching starts and parts of identifiers it's better to just use them.

This issue is really important as it prevents completely valid code from running.

### DIFF
--- a/pebble/src/main/java/com/mitchellbosecke/pebble/lexer/LexerImpl.java
+++ b/pebble/src/main/java/com/mitchellbosecke/pebble/lexer/LexerImpl.java
@@ -89,7 +89,7 @@ public final class LexerImpl implements Lexer {
   /**
    * Static regular expressions for names, numbers, and punctuation.
    */
-  private static final Pattern REGEX_NAME = Pattern.compile("^[a-zA-Z_][a-zA-Z0-9_]*");
+  private static final Pattern REGEX_NAME = Pattern.compile("^\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*");
 
   private static final Pattern REGEX_LONG = Pattern.compile("^[0-9]+L");
 
--- a/pebble/src/main/java/com/mitchellbosecke/pebble/lexer/LexerImpl.java
+++ b/pebble/src/main/java/com/mitchellbosecke/pebble/lexer/LexerImpl.java
@@ -89,7 +89,7 @@ public final class LexerImpl implements Lexer {
   /**
    * Static regular expressions for names, numbers, and punctuation.
    */
-  private static final Pattern REGEX_NAME = Pattern.compile("^\\p{javaJavaIdentifierStart}\\p{javaJavaIdentifierPart}*");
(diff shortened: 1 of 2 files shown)


ebussieres:
I don't really understand why you would have methods or fields with characters like éêȩ ??? I speak French and i NEVER use those characters.

Also, there's no unit test in your PR

objectProgrammer:
Personal preference. Not every code needs to be in English :) Also, if you want to call existing methods or use existing fields which use non-English then you need to make messy wrappers to call from your pebble code.

What's to be unit tested? Apologies, but I haven't written many unit tests, not sure what I should be testing (it's just a regex change). If you think there might be an error with the regex \\p{} pattern itself then I doubt, because it's been in Java for so long, any issues would've been found.

ebussieres:
You can take a look at any test classes. Ex. here : https://github.com/PebbleTemplates/pebble/blob/master/pebble/src/test/java/com/mitchellbosecke/pebble/CoreTagsTest.java

Every modification needs to be covered by unit tests.

In your case, add tests to demonstrate what are the new accepted characters   (éèê etc.)

objectProgrammer:
@ebussieres Okay, I guess currency symbols are not needed but the only punctuation I know of is _ which is already supported.

If it's okay for you, adding support for all Latin characters instead of all identifier characters would be fine (`\\p{IsLatin}`).

I'm not sure if it is a problem because I doubt pebble supports it anyways but IsLatin does not work on Android, for whatever reason they call it "Latin" only. If it is a problem I can add a check or just use a normal regex pattern with ranges other than just A-Z.

I'll try to look into tests next week.

objectProgrammer:
@ebussieres How would you prefer it done? This is an important issue.
- IsLatin, if Android is supported then static check for it once
- More ranges (there are some to select many diacritic-characters at once)

objectProgrammer:
I found this closed issue #414, which is somewhat related. I believe \\p{IsLetter} along with \\p{IsDigit} and _ is better, as it supports more scripts and covers all usages I can think of.

objectProgrammer:
@ebussieres I've added tests as well, if there aren't any more problems then please merge :)

ebussieres:
Your unit tests are failing. Please fix them

objectProgrammer:
Huh.

objectProgrammer:
How on earth am I going to know which template it is talking about, it just says "(template.peb:3)"! There are a ton of template.peb, I need to know which /testxx directory....

objectProgrammer:
The problem seems to be with the tests? src/test/resources/templates/embed/test15/template.peb uses src/test/resources/templates/embed/test15/template.base.peb but does not put the content in blocks.

objectProgrammer:
Exactly. The error happens with a clean clone as well, might me my java version?

objectProgrammer:
Nevermind all this. It was just that I moved the file without adding imports.

Issue #520, opened 2020-06-26
Is it possible to configure which core functionality is loaded per default?


pepperbob:
We're using pebble templates to generate short reference numbers that are usually 10 to 30 chars long. The way these reference numbers are composed must be highly configurable so we basically expose the required pebble templates to (administrative) users.

As not all functionality provided by pepple is useful for these cases I wonder if there's a way to unload core extensions to limit available functionality as much as possible for security reasons and some control what a user can actually do with the templates.

Any thoughts on this is much appreciated!

ebussieres:
As of now, Pebble provides the following extension that cannot be removed 

- CoreExtension
- EscaperExtension
- i18nExtension
- AttributeResolverExtension

 It would need some adjustment in PebbleEngine for that. However, it is possible to override operators with the `allowOverrideCoreOperators` flag.

If there's a use case for that, I think that you can open a PR to allow the user to enable/disable some extensions

pepperbob:
Sorry for coming back so late on this topic... 

So when using the `PebbleEngine.Builder` a ExtensionRegistry is created which registers amongst others the `CoreExtension` which itself then provides tags, filters, etc. that will be used in the templates.

As an example: I need to restrict my the setup so that some tags cannot be used.

Is your suggestion to control the creation of `ExtensionRegistry` (PebbleEngine:621) by providing a factory for it through which I would be able to register my own Version of `CoreExtension` - which might be an inherited version overwriting `getTokenParsers`?

This sounds feasible, hopefully it won't dilute pebbles configuration-API too much. I'll take care to provide a MR for this and we'll see.

PR #552, opened 2020-11-29
issue #520: provide option to customize stock extensions

pepperbob:
This is a draft for #520 to allow for a customization of the default extensions.

- refactored out the extension loading code from the `PebbleEngine.Builder` into the `ExtensionRegistryFactory` 
- provide a `ExtensionCustomizer` which basically allows to access extensions before loading

### DIFF
--- a/docs/src/orchid/resources/wiki/guide/customize-defaults.md
+++ b/docs/src/orchid/resources/wiki/guide/customize-defaults.md
@@ -1,0 +1,37 @@
+---
+---
+
+Pebble comes with a rich set of built-in tags and filters that will help you render your templates into  websites and other documents with ease. However, imagine a more specific use-case where the templates are not entirely under your control.
+
+In these cases it might be advised to consider stripping-down Pebbles' built-in functionality that may otherwise introduce security-concers regarding the integrity and stability of your application.
+
+### Opt-Out using ExtensionCustomizer
+
+The `ExtensionCustomizer` base class can be used to gain access to the default functionality before it is loaded into Pebbles template engine. Overwrite methods to get hold on provided default-functionality and modify whatever should be available for the template engine.
+
+The following example removes the `ForTokenParser`, i.e. the ability to parse `{% for %}{{ ... }}{% endfor %}` constructs:
+
+```java
+class ExampleOptOuts extends ExtensionCustomizer {
+
+  public ExampleOptOuts(Extension ext) {
+    super(ext);
+  }
+
+  @Override
+  public List<TokenParser> getTokenParsers() {
+    List<TokenParser> tokenParsers = Optional.ofNullable(super.getTokenParsers())
+                                        .map(ArrayList::new).orElseGet(ArrayList::new);
+      
+    tokenParsers.removeIf(x -> x instanceof ForTokenParser);
+    return tokenParsers;
+  }
+
+}
+```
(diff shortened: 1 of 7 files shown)


pepperbob:
So I came up with "ExtensionCustomizer" to allow filtering out or change whatever functionality is provided by an extension. It currently only targets the baked in extensions - not sure if it would make much sense on the user provided extensions anyway. 

On the one hand it looks like it leaves the configuration API completely intact and doesn't degrade things like `allowOverrideCoreOperators` to an implementation detail.

On the other hand I'm not too happy with the current version as now other implementation details, e.g. how core functionality is split over extension classes (core, i18n, escaper) kind of leaks out as you have to know the extension class to apply the customizer.

Of course the whole target class thing can be dropped and there could be just one single customizer that is wrapped around every extension before it is loaded ... while thinking about it - i think this is would would provide the cleanest API and won't break things for users if stuff is re-organized internally at some point.. I'll rework it.

Feedback is much appreciated!

pebble/src/test/java/com/mitchellbosecke/pebble/extension/ExtensionCustomizerTest.java
```diff
@@ -37,15 +39,16 @@ void upperFilterCannotBeUsed() throws IOException {
...
 
-    public RemoveUpper(Extension core) {
+    public RemoveUpperCustomizer(Extension core) {
       super(core);
     }
 
     @Override
     public Map<String, Filter> getFilters() {
-      Map<String, Filter> filters = new HashMap<>(super.getFilters());
+      Map<String, Filter> filters = Optional.ofNullable(super.getFilters()).map(HashMap::new)
```
pepperbob, line 50: just curious: is there a reason why null is acceptable over an empty list if filters and others are not provided?

ebussieres:
Can you provide some documentation for this new functionality ?

pepperbob:
@ebussieres Ok, I added further JavaDoc - is that okay so far?

ebussieres:
I mean add some documentation for pebble website. Doc is here : https://github.com/PebbleTemplates/pebble/tree/master/docs/src/orchid/resources

pepperbob:
Oh sure, I'll check this out!

docs/src/orchid/resources/wiki/guide/customize-defaults.md
```diff
@@ -0,0 +1,37 @@
...
+In these cases it might be advised to consider stripping-down Pebbles' built-in functionality that may otherwise introduce security-concers regarding the integrity and stability of your application.
+
+### Opt-Out using ExtensionCustomizer
+
+The `ExtensionCustomizer` base class can be used to gain access to the default functionality before it is loaded into Pebbles template engine. Overwrite methods to get hold on provided default-functionality and modify whatever should be available for the template engine.
+
+The following example removes the `ForTokenParser`, i.e. the ability to parse `{% for %}{{ ... }}{% endfor %}` constructs:
+
+```java
+class ExampleOptOuts extends ExntensionCustomizer {
```
ebussieres, line 15: there's a typo in `ExntensionCustomizer`

pepperbob: :facepalm: fixed it :)

PR #564, opened 2020-12-12
Ability to check for presence of variables from EvaluationContext

objectProgrammer:
Currently getVariable returns null for non-present variables, so you can't know if the value is null or if it doesn't exist.
This pull request adds a containsVariable method to fix this.

### DIFF
--- a/pebble/src/main/java/com/mitchellbosecke/pebble/template/EvaluationContext.java
+++ b/pebble/src/main/java/com/mitchellbosecke/pebble/template/EvaluationContext.java
@@ -12,4 +12,6 @@ public interface EvaluationContext {
   Locale getLocale();
 
   Object getVariable(String key);
+
+  boolean containsVariable(String key);
 }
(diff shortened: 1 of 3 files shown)


ebussieres:
Please add a unit test

objectProgrammer:
@ebussieres Do you know how I can get the EvaluationContext without an extension? Needed for the test.

objectProgrammer:
I really have no idea how to get it. Besides, this just adds 2 methods that aren't even called by the core itself, are tests really needed?

objectProgrammer:
@ebussieres I've looked a lot but there is no way to get EvalutationContext?

objectProgrammer:
Well, as I got no suggestion on how to get it I used reflection. If it's bad then I'm afraid this can't be merged because there is no other way that I know of. And if you don't merge then I don't care so much because I found that you can cast to EvaluationContextImpl and use getScopeChain to check for contains anyways.

objectProgrammer:
I doubt this will get merged.

Issue #568, opened 2020-12-31
Cannot use sort filter right after split filter


ezanaga1:
I would like to split and sort a string directly in template but the following code:
`{{ 'q,g,s,c,w' | split(',') | sort }}`

Raises the exception:
> java.lang.ClassCastException: class [Ljava.lang.String; cannot be cast to class java.util.List ([Ljava.lang.String; and java.util.List are in module java.base of loader 'bootstrap')
> 	at com.mitchellbosecke.pebble.extension.core.SortFilter.apply(SortFilter.java:32)
> 	at com.mitchellbosecke.pebble.extension.core.SortFilter.apply(SortFilter.java:18)
> 	at com.mitchellbosecke.pebble.node.expression.FilterExpression.evaluate(FilterExpression.java:72)
> 	at com.mitchellbosecke.pebble.node.expression.FilterExpression.evaluate(FilterExpression.java:65)
> ....

This because split filter returns the result of String.split(), a String[], and sort filter wants a List in input.
IMHO, is better use Collection-like object instead of plain Java arrays, so it is possible to wrap String.split() result in a List, like Arrays.asList(String.split()) ?

ebussieres:
IMO, the solution should be to modify sort filter to support an array or a List

PR #569, opened 2021-01-01
#568: Support array of List in sort filter

ebussieres:


objectProgrammer:
Why not Arrays.sort for arrays and Collections.sort for Collections? Seems redundant to wrap it in Arrays.asList and then treat it as a Collection.

Issue #573, opened 2021-02-01
Adding a filter to hash a string to SHA256


silviuvergoti:
I used more than once a filter to hash a string to SHA256. I could add it to the core extension, if other users could benefit from it.

ebussieres:
Sounds good to me, just make a PR and I'll review it

Issue #572, opened 2021-02-01
Adding a filter for base64 encoding


silviuvergoti:
One of the filters that I used multiple times was a base64-encoding of the original string.  I was wondering if I could add the filter to the core extension?

ebussieres:
Sounds like a good idea to me

PR #574, opened 2021-02-03
New base64encode and sha256 filters

silviuvergoti:
Adding new filters.
Fixes issues  #573 and #572.

### DIFF
--- a/pebble/src/main/java/com/mitchellbosecke/pebble/extension/core/Base64EncoderFilter.java
+++ b/pebble/src/main/java/com/mitchellbosecke/pebble/extension/core/Base64EncoderFilter.java
@@ -1,0 +1,37 @@
+package com.mitchellbosecke.pebble.extension.core;
+
+import com.mitchellbosecke.pebble.error.PebbleException;
+import com.mitchellbosecke.pebble.extension.Filter;
+import com.mitchellbosecke.pebble.template.EvaluationContext;
+import com.mitchellbosecke.pebble.template.PebbleTemplate;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * This class implements the 'base64encode' filter.
+ *
+ * @author Silviu Vergoti
+ */
+public class Base64EncoderFilter implements Filter {
+
+  public static final String FILTER_NAME = "base64encode";
+
+  @Override
+  public List<String> getArgumentNames() {
+    return null;
+  }
+
+  @Override
+  public Object apply(Object input, Map<String, Object> args, PebbleTemplate self,
+      EvaluationContext context, int lineNumber) throws PebbleException {
+    if (input == null) {
+      return null;
+    }
+
+    return Base64.getEncoder().encodeToString(input.toString().getBytes(StandardCharsets.UTF_8));
+  }
+
+}
--- a/pebble/src/main/java/com/mitchellbosecke/pebble/extension/core/Base64EncoderFilter.java
+++ b/pebble/src/main/java/com/mitchellbosecke/pebble/extension/core/Base64EncoderFilter.java
@@ -17,7 +17,7 @@ import java.util.Map;
  */
 public class Base64EncoderFilter implements Filter {
 
-  public static final String FILTER_NAME = "base64encode";
+  public static final String FILTER_NAME = "base64";
 
   @Override
   public List<String> getArgumentNames() {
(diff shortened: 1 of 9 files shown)


objectProgrammer:
Perhaps base64 is common enough to be in the core, but if SHA256 gets added then suddenly we could end up with MD5, SHA-1, SHA-2, SHA-3 and more. I feel like there are too many hashing functions for any of them to be included in the core.

Just curious, what's the use for having a hashing filter?

And base64encode sounds too long, maybe just `base64`?

silviuvergoti:
Sure, I'll shorten it. I thought that a simple `base64` would not be clear enough if it concerns encoding or decoding.  
Regarding the hashing filters, there are a lot of web services requiring hashing of PII (emails, ids, phones). If someone needs it, the filter could be changed to something like `{{ originalstring | hash('SHA256') }}` and it could support many more hash algorithms.

pebble/src/main/java/com/mitchellbosecke/pebble/extension/core/CoreExtension.java
```diff
@@ -149,6 +149,8 @@
     filters.put(ReplaceFilter.FILTER_NAME, new ReplaceFilter());
     filters.put(MergeFilter.FILTER_NAME, new MergeFilter());
     filters.put(SplitFilter.FILTER_NAME, new SplitFilter());
+    filters.put(Base64EncoderFilter.FILTER_NAME, new Base64EncoderFilter());
```
ebussieres, line 152: Should we provide a base64 decoder too ?

pebble/src/test/java/com/mitchellbosecke/pebble/CoreFiltersTest.java
```diff
@@ -1329,6 +1329,57 @@ void testReplaceFilterInTemplate() throws PebbleException, IOException {
     assertEquals("I like foo and bar.", writer.toString());
   }
 
+  /**
+   * Tests {@link com.mitchellbosecke.pebble.extension.core.Base64EncoderFilter} if the base64 encoding filter is working for a string value, a string constant, null.
+   */
+  @Test
+  void testBase64EncoderFilterInTemplate() throws PebbleException, IOException {
```
ebussieres, line 1336: Add a unit test with null as value passed to base64 filter

pebble/src/main/java/com/mitchellbosecke/pebble/extension/core/CoreExtension.java
silviuvergoti: @ebussieres, decoding a base64 string will return a byte array. If it's a binary object, reconstructing it might not be feasible. Do you suggest a return value type of `byte[]`? The encoder was mainly targeted to strings with special characters.

ebussieres:
```
byte[] decodedBytes = Base64.getDecoder().decode(encodedString);
String decodedString = new String(decodedBytes);
```

pebble/src/main/java/com/mitchellbosecke/pebble/extension/core/Base64EncoderFilter.java
```diff
@@ -0,0 +1,37 @@
...
+import java.util.Base64;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * This class implements the 'base64encode' filter.
+ *
+ * @author Silviu Vergoti
+ */
+public class Base64EncoderFilter implements Filter {
```
ebussieres, line 18: Can you add some documentation for the new filters too ? 

https://github.com/PebbleTemplates/pebble/tree/master/docs/src/orchid/resources

pebble/src/main/java/com/mitchellbosecke/pebble/extension/core/CoreExtension.java
objectProgrammer: What's the use case? I mean base64 encoding can be useful for putting in a link or showing a raw image but decoding seems pretty useless, if it's data from another server I'd expect it to be decoded before being provided to the template. Besides, new String(...) is loose and will probably cause problems on Windows machines, StandardCharsets.UTF_8 should be supplied.

silviuvergoti: I'll add the decoder and it will try to return an UTF-8 String. If there's an exception, we'll throw it. No object casting will be done.

pebble/src/main/java/com/mitchellbosecke/pebble/extension/core/Base64EncoderFilter.java
silviuvergoti: Sure.

PR #596, opened 2021-06-16
Fix #595 : Numeric operator overflow

### LINKED ISSUE #595: Long Comparisons Fail Due to Double Conversion
## Problem

We came across an issue today and believe I have figured out why it occurs. When using pebble with id values that are longs we noticed odd behavior where pebble considered the values `325055142682428416` and `325055142682428417` to be equal. You can recreate this issue with any large longs.

After reviewing the code we realized that all number comparisons in Pebble are converted to Double values which we believe is the issue. Please see the following lines for reference:

The line that calls `wideningConversionBinaryComparison` from a comparison where the type is `Number`:
https://github.com/PebbleTemplates/pebble/blob/aa43888a87901ea892258c02cee781d183d4a06b/pebble/src/main/java/com/mitchellbosecke/pebble/utils/OperatorUtils.java#L69

The line in `wideningConversionBinaryComparison` that actually performs the double comparison:
https://github.com/PebbleTemplates/pebble/blob/aa43888a87901ea892258c02cee781d183d4a06b/pebble/src/main/java/com/mitchellbosecke/pebble/utils/OperatorUtils.java#L191

## Solution

I would recommend adding a type check for `Number` to also see if the instance also implements `Comparable` as this should work for many numerics reliably without conversion. This would also give you support for `BigDecimal` and `BigInteger` which we also use a lot for their reliability in financial math. The downside of `Comparable` is it can be tricky if you have to handle the case of comparing two numbers of different types, where both implement `Comparable<T>` of their own type, but a class cast exception would be thrown with `double.compareTo(long)` for example.

Here to assist as needed if we can contribute, but this is obviously a sensitive low level operator for Pebble, so requires your input on implementation I believe. :+1:

RedefineMike:
Initial pull request that just fixes the comparison operators and could be greatly improved by:
- Merging type specific checks with the already type specific arithmetic operators, applying this fix to all Numeric Binary Expressions.
- More careful consideration of converting numbers of different types, converting the operands to types that can always appropriately handle both numeric values.


RedefineMike:
If a maintainer could review everything, I've completed a significant change to the Binary Comparison and Operation methods in `OperatorUtils` that should handle all numeric types effectively. When the two operands are of the same class I simplified the type checks.

In all cases, the code only converts a number between types when necessary. For operands of differing types, it converts them to the highest complexity / scale type of the two.

As a bonus, I also added BigInteger support.

pebble/src/main/java/io/pebbletemplates/pebble/utils/OperatorUtils.java
```diff
@@ -151,29 +152,56 @@ private static Object wideningConversionBinaryOperation(Object op1, Object op2,
...
+        return bigIntegerOperation((BigInteger) num1, (BigInteger) num2, operation);
+      }
     }
 
-    if (num1 instanceof Double || num2 instanceof Double) {
-      return doubleOperation(num1.doubleValue(), num2.doubleValue(), operation);
+    /*
+     * Convert and compare operands based on the complexity / scale of the values.
+     */
+    if (num1 instanceof BigDecimal) {
```
ebussieres, line 177: Need to add tests when both operands are not the same type

ebussieres: Should also add test for numeric operation (add, subtract, multiply, etc.)

ebussieres:
Great PR, I'll gladly accept it when test coverage is gonna be better. Can you adjust it ?

thecoden:
Hi @ebussieres , I somehow missed your comments but will review and see if I can find some time to add more test cases.

PR #611, opened 2021-12-06
Added option to allow access to entire ScopeChain from within macros. Default behavior is as before

SharkFourSix:
* Added configuration variable ``globalScoping`` to ``pebble/src/main/java/com/mitchellbosecke/pebble/PebbleEngine.java:66`` which effectively allows macros to be able to access variables in the entire scope chain.
* Added unit test case for the same.

To enable global scoping:
```java
new PebbleEngine.Builder().globalScoping(true)...
```

Also, excuse the IntelliJ formatting. Perhaps adding a **.editorconfig** will be nice.


ebussieres:
We should also disable the special "_context" variables when globalScoping is set to true

SharkFourSix:
Do you mean here:

https://github.com/PebbleTemplates/pebble/blob/f1f33368cfbfab7c9578b45ace7276af56cca886/pebble/src/main/java/io/pebbletemplates/pebble/template/PebbleTemplateImpl.java#L187

ebussieres:
Yeah exactly, and here too 

https://github.com/PebbleTemplates/pebble/blob/master/pebble-spring/pebble-spring6/src/main/java/io/pebbletemplates/spring/extension/function/bindingresult/BaseBindingResultFunction.java#L44

https://github.com/PebbleTemplates/pebble/blob/master/pebble-spring/pebble-spring5/src/main/java/io/pebbletemplates/spring/extension/function/bindingresult/BaseBindingResultFunction.java#L44

ebussieres:
Will close since there has been no activity for a long time

PR #616, opened 2022-02-08
Improve MAVEN build Performance

xiayingfeng:
That report generation takes time, slowing down the overall build. Reports are definitely useful, but do you need them every time you run the build. We can conditionally disable generating test reports by setting `<disableXmlReport>true<disableXmlReport>`. If you need to generate reports, just add `-DcloseTestReports=false` to the end of mvn build command.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -20,6 +20,7 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <java.version>1.8</java.version>
+  	<closeTestReports>true</closeTestReports>
   </properties>
 
   <licenses>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -20,6 +20,7 @@
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <java.version>1.8</java.version>
+  	<closeTestReports>true</closeTestReports>
```
crummy, line 23: FYI it looks like you've changed from spaces to tabs in both lines you changed

PR #638, opened 2022-12-09
i18n-alias - adding i18 function alias with `messages` bundle as default

ilyas-gx:
Most of the time there is only one bundle - `messages`, so this shortcut/alias is really useful for that cases.


ebussieres:
Instead of creating a new function with a funky name (Just 't'), we should modify the i18n function to use a default value if no value for bundle are found.

ilyas-gx:
@ebussieres , this "funky" name is well known and widely used by dev community across many projects.

But it's up to you, guys. I already have this code in my own projects so I'm just sharing.

PR #650, opened 2023-03-24
Fix the null pointer exception on the replace filter

fabriceci:
There is a check missing before doing toString() on the input.


ebussieres:
Can you add a unit test for this case please ?

Thank you for your contribution

fabriceci:
Of course, it's done.

PR #654, opened 2023-07-02
Support Beans#get with Class

zack-harrison:
Instead of always calling toString, allow getting a bean by Class. This is a rare use case, but we would like to use it for example in an Extension on the Java side which needs to get beans from Spring context.


ebussieres:
Can you add some unit tests ?

PR #663, opened 2023-11-13
Generate OSGi metadata into pebble jar to make it OSGi compliant

HannesWell:
This allows to use the pebble jar in an OSGi runtime, like Eclipse, out of the box.

The version ranges for the optional dependencies `com.github.ben-manes.caffeine:caffeine`, `javax.servlet` and `jakarta.servlet` have been widened manually because there are newer versions available than the one this library depends on.

If you are interested in updating any of these dependencies (and especially migrate `javax.servlet:servlet-api` to `javax.servlet:javax.servlet-api` or even `jakarta.servlet:jakarta.servlet-api:4.0.2` (which provides the `javax.servlet` packages) I can provide a separate PR or do it in this one.

The OSGi compliant MANIFEST.MF is only generated for the main `pebble` artifact.
If you think this is also relevant for the spring artifacts I can extend the change, but I would not be able to test the result.

### DIFF
--- a/pebble/pom.xml
+++ b/pebble/pom.xml
@@ -90,12 +90,36 @@
   <build>
     <plugins>
       <plugin>
+        <groupId>biz.aQute.bnd</groupId>
+        <artifactId>bnd-maven-plugin</artifactId>
+        <version>7.0.0</version>
+        <executions>
+          <execution>
+            <id>generate-osgi-manifest</id>
+            <goals>
+              <goal>bnd-process</goal>
+            </goals>
+            <configuration>
+              <bnd><![CDATA[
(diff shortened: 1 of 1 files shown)


HannesWell:
@ebussieres could you please have a look at this?

HannesWell:
Thank you @ebussieres for accepting this.
Do you already have some idea when the next release will happen?

ebussieres:
I'll make one today

HannesWell:
> I'll make one today

Awesome. Thanks again!

ebussieres:
@HannesWell : Release has been made.

PR #671, opened 2024-03-02
add ConfigurableExtensionCustomizer

freshchen:
we needs to dynamically open different feature according to different scenarios.
After adding this, most default feature choose can be done through config files without writing code

### DIFF
--- a/docs/src/orchid/resources/wiki/guide/customize-defaults.md
+++ b/docs/src/orchid/resources/wiki/guide/customize-defaults.md
@@ -44,7 +44,7 @@ For example of use, see below:
 
 ```java
 PebbleEngine engine = new PebbleEngine.Builder()
-        .registerExtensionCustomizer(DisallowExtensionCustomizerBuilder.builder()
+        .registerExtensionCustomizer(new DisallowExtensionCustomizerBuilder()
                 .disallowedTokenParserTags(singletonList("flush"))
                 .disallowedFunctionKeys(singletonList("max"))
                 .disallowedFilterKeys(singletonList("upper"))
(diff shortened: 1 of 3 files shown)


pebble/src/main/java/io/pebbletemplates/pebble/extension/core/ConfigurableExtensionCustomizer.java
```diff
@@ -0,0 +1,147 @@
...
+import java.util.Objects;
+import java.util.Optional;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+/**
+ * @author freshchen
+ * @since 2024/3/2
+ */
+public class ConfigurableExtensionCustomizer {
```
ebussieres, line 25: Maybe rename this class to DisallowExtensionCustomizerBuilder

pebble/src/main/java/io/pebbletemplates/pebble/extension/core/ConfigurableExtensionCustomizer.java
```diff
@@ -0,0 +1,147 @@
...
+  private Collection<String> disallowedUnaryOperatorSymbols;
+
+  private Collection<String> disallowedTestKeys;
+
+
+  public static ConfigurableExtensionCustomizer builder() {
+    return new ConfigurableExtensionCustomizer();
+  }
+
+  public ConfigurableExtensionCustomizer setDisallowedFunctionKeys(Collection<String> disallowedFunctionKeys) {
```
ebussieres, line 44: remove set prefix from all method name of the builder

ebussieres (changes requested):
Please modify documentation (docs/src/orchid/resources/wiki/guide/customize-defaults.md) accordingly too

pebble/src/main/java/io/pebbletemplates/pebble/extension/core/ConfigurableExtensionCustomizer.java
freshchen: good suggestion

Issue #625, opened 2022-09-15
Vulnerability CVE-2022-37767


pelece:
Hello,

since this morning security checks in our projects are reporting new critical vulnerability in the current pebble version 3.1.5:
NVD: https://nvd.nist.gov/vuln/detail/CVE-2022-37767
Original report: https://github.com/Y4tacker/Web-Security/issues/3

Any short-term workaround to mitigate the vulnerability?

slandelle:
IMHO, using deny lists is an error and pebble should switch to using allow lists instead: the end-user should be responsible of deciding which classes are safe and which are not.

@ebussieres WDYT?

ebussieres:
We already had a conversation on this https://github.com/PebbleTemplates/pebble/pull/494#issuecomment-629691422

chaitu0292:
@ebussieres Sorry I couldn't get to any conclusion looking at #494.
Do you mean the reported CVE is already addressed? or Is it in your pipeline to resolve?

rumfuddle:
+1
This CVE is breaking the build in any corporate environment and it's a bit sad that it's not taken serious.

slandelle:
Disclaimer: I'm just an occasional contributor here

@chaitu0292 No, this CVE is not addressed. IMHO, the blacklist approach that was picked cannot possibly cover all possible security leaks and only a whitelist approach where the user explicitly list what's needed in his templates can work. But maybe I'm missing something.

@rumfuddle Then, corporate users who depend on this library should not stay idle and wait for some open-source volunteers to tackle this issue on their free time. They should contribute, either in manpower or financially.

rumfuddle:
> @rumfuddle Then, corporate users who depend on this library should not stay idle and wait for some open-source volunteers to tackle this issue on their free time. They should contribute, either in manpower or financially.

Fair point, but if the PR consists of just an upgrade of the dependency containing the CVE, will you approve?

ebussieres:
@rumfuddle No it's not just an upgrade of a dependency. You need to define a MethodAccessValidator [implementation](https://github.com/PebbleTemplates/pebble/blob/master/pebble/src/main/java/com/mitchellbosecke/pebble/attributes/methodaccess/MethodAccessValidator.java)

Then
If you use pebble-spring:
- Define a bean which implements `MethodAccessValidator`  

```
    @Bean
    public MethodAccessValidator methodAccessValidator() {
      return new BlacklistMethodAccessValidator(); //Provide your custom implementation
[...]

PR #672, opened 2024-03-04
add maxLoopIterationNumber

freshchen:
Prevent malicious templates from draining CPU resources


freshchen:
Our scenario is LLM Agent.
In order to make it more convenient for everyone, try to combine business and LLM capabilities. we provides SaaS services
Users complete simple processing of registered API tools through the template engine.

So we cannot limit the template content

ebussieres:
If you don't control your templates, you're gonna be at risk. Please read this : https://github.com/PebbleTemplates/pebble/issues/625

freshchen:
Thanks for notifying this very useful issue. 

In addition, can we refer to the groovy sandbox on the basis of MethodAccessValidator to conduct security risk control on the entire execution process of the template.

Personally, I prefer the implementation of whitelisting.

PR #695, opened 2025-07-22
NaN must return false

serhii-yatskovskyi:


ebussieres:
Please add a unit test

PR #741, opened 2025-12-07
Enhancement: Support dynamic named arguments

### LINKED ISSUE #740: Enhancement: Support dynamic named arguments
I do have a filter `queryString` which by default generates a queryString like `/api/library?q=string&isbn=string`

default usage:
```
{{operation | queryString}}
```

custom usage:
```
{{operation | queryString(q="Dune")}}
```

This last example should generate just: `/api/library?q=Dune`

My filter returns `List.of()` on argument names, but still names are validated.

jknack:
- support dynamic/unknown argument names
- fix #740


ebussieres:
Can you add some documentation about how to use dynamic named arguments ? Documentation about named arguments are here: 

- https://github.com/PebbleTemplates/pebble/blob/master/docs/src/orchid/resources/wiki/guide/basic-usage.md?plain=1#L243
- https://github.com/PebbleTemplates/pebble/blob/master/docs/src/orchid/resources/wiki/guide/extending-pebble.md?plain=1#L21

ebussieres:
What I don't like about this solution is that we use a shortcut. empty list vs null have two different behaviour. I think we should refactor getArgumentMap instead of making such shortcut

PR #768, opened 2026-02-09
Highlight current page from sidebar on wiki pages

dashtinejad:
When browsing wiki pages, it is inconvenient to repeatedly reopen the accordion to navigate between pages.

This PR ensures that the currently active accordion in the WIKI section remains open and highlights the selected page.

For instance, while browsing this page: https://pebbletemplates.io/wiki/function/blockFunction/

The "FUNCTION" accordian will be opened and "BLOCK" will be highlighted automaticaly.

image

### DIFF
--- a/docs/src/orchid/resources/templates/includes/sidebar.peb
+++ b/docs/src/orchid/resources/templates/includes/sidebar.peb
@@ -58,28 +58,3 @@
 
     </div>
 </div>
-
-<script>
-(function ($) {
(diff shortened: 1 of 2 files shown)


dashtinejad:
@ebussieres I wasn’t able to get Orchid running. I verified the functionality by testing it manually in the browser, but it would be good to also verify it through Orchid to ensure everything works as expected.

PR #816, opened 2026-07-02
Reduce allocations and duplication in attribute resolution

Nafis8709:
## Problem

`DefaultAttributeResolver` and `MemberCacheUtils` sit on the attribute-resolution path that
runs on every `{{ object.property }}` access during rendering, so small inefficiencies here
are multiplied across every attribute in every template. Four issues live in this path:

- `getArgumentTypes()` returns `new Class<?>[0]` on every no-argument access — the common
  case (`{{ user.name }}`) — allocating a throwaway empty array each time.
- `MemberCacheKey` is a non-static inner class that never uses its enclosing instance, so
  every key stored in the cache silently holds a reference to the outer `MemberCacheUtils`.
- `reflect()` contains three near-identical blocks trying the `get` / `is` / `has` prefixes.
- `findMethod()` mixes the "perfect match" and "greedy match" strategies in one long method.

## Change

- **Hot-path allocation:** `getArgumentTypes()` returns a shared `EMPTY_ARGUMENT_TYPES`
  constant for the no-argument case instead of allocating each time.
- **Hidden reference:** `MemberCacheKey` is now `static`, removing the implicit outer
  reference from every cached key.
- **Duplication:** the three `get`/`is`/`has` lookups collapse into a single loop over a
  prefix array, tried in the same order as before.
- **Long method:** `findMethod()` is split into `findPerfectMatch()` and `findGreedyMatch()`.
  `verifyUnsafeMethod()` is still called at the same point, so matching and access validation
  are unchanged.

**Non-breaking:** the public API of `DefaultAttributeResolver` is unchanged, and method
matching order and results are identical. These are pure refactorings.

## Tests

- Full test suite green: `mvn clean test -pl pebble -am` — <FILL IN> tests, 0 failures, 0 errors.
- No test changes required, since behaviour is unchanged.

### DIFF
--- a/pebble/src/main/java/io/pebbletemplates/pebble/attributes/DefaultAttributeResolver.java
+++ b/pebble/src/main/java/io/pebbletemplates/pebble/attributes/DefaultAttributeResolver.java
@@ -17,14 +17,21 @@ public class DefaultAttributeResolver implements AttributeResolver {
 
   private final MemberCacheUtils memberCacheUtils = new MemberCacheUtils();
 
+  // REFACTOR (hidden allocation on the hot path): plain property access such as
+  // {{ user.name }} passes no arguments, so getArgumentTypes previously allocated
+  // a brand-new empty Class<?>[] on every single attribute access. A shared,
+  // immutable empty array is reused instead. An empty array is stateless, so
+  // sharing one instance is safe.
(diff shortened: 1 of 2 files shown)


ebussieres:
AI stuff I presume ?

